PR #3, opened 2019-04-01
Introduce source file output feature

dmlloyd:


FroMage:
This looks great. How do I get the same with files generated with ASM? :)

dmlloyd:
That'd be a little bit more of a challenge. :)  Possible though!

dmlloyd:
Any further thoughts?  Shall I merge?

FroMage:
I would use that, and my IDE does not have a disassembler by default.

PR #8, opened 2019-05-28
Fix source info bytecode problems

dmlloyd:


dmlloyd:
Without this commit, the `zig` files are written but not referenced from the bytecode so debuggers won't follow it along and exception stack traces won't reference the file name/line number.

dmlloyd:
Hang on, there's a second problem.

dmlloyd:
Fixed

dmlloyd:
No objections, merged

PR #12, opened 2019-06-24
Allow class bytes to be written separately from close

### LINKED ISSUE #11: Allow class bytes to be written after the class creator is closed
This lays groundwork for fixing quarkusio/quarkus#2923.

dmlloyd:
Fixes #11


dmlloyd:
If nobody raises an issue with this by Monday, I'm going to go ahead and merge it.

PR #23, opened 2020-01-13
Add convenience newArray method

geoand:
This is similar to what `readArrayValue` does


src/main/java/io/quarkus/gizmo/BytecodeCreator.java
```diff
@@ -152,6 +152,10 @@ default ResultHandle newArray(Class type, ResultHandle length) {
         return newArray(type.getName(), length);
     }
 
+    default ResultHandle newArray(Class type, int length) {
```
stuartwdouglas: This should be Class<?>, and we should also add the String based version as well for consistency.

geoand: Done

PR #26, opened 2020-01-28
Support for array annotation values

gytis:


gsmet:
@stuartwdouglas if you can have a look at this one and that one https://github.com/quarkusio/gizmo/pull/25 , I can cut a new Gizmo release tomorrow.

src/main/java/io/quarkus/gizmo/AnnotationUtils.java
```diff
@@ -0,0 +1,23 @@
+package io.quarkus.gizmo;
+
+import org.jboss.jandex.AnnotationValue;
+import org.objectweb.asm.AnnotationVisitor;
+
+public final class AnnotationUtils {
```
stuartwdouglas: This should probably be package private.

gytis: @stuartwdouglas done. I made both class and method package-private. Is that OK or should only be a class?

stuartwdouglas: Perfect.

PR #36, opened 2020-03-30
Bump to ASM 8

geoand:


geoand:
This is gizmo :)

gastaldi:
image

gsmet:
So my question about ByteBuddy stays. We need a coordinated upgrade and we don't want to upgrade Gizmo until we are sure we can upgrade things in Quarkus itself so that means we need to be sure ByteBuddy (and Hibernate ORM) works well with ASM8.

But it's good to have a feedback from CI in any case and have a green light for Gizmo.

geoand:
Yeah, that is obviously the case, but this would need to be merged in any case :)

gsmet:
No. I would say we need to merge it once we are sure we want it.

For now, I'm not sure until we check ByteBuddy and ORM.

gsmet:
Latest ByteBuddy is still using 7.3.1 so we will have to wait a bit (usually, they upgrade to the latest very soon after an ASM release but circumstances are a bit exceptional so we'll see).

geoand:
Yeah, ByteBuddy 1.10.9 is on ASM 7.3.1: https://mvnrepository.com/artifact/net.bytebuddy/byte-buddy-dep/1.10.9

FWIW, I am not advocating releasing Gizmo and updating Quarkus to use it, I am advocating we merge the PR only, although I do get where you are coming from. You're guarding against the scenario where we need to release Gizmo to fix some bug - that's understandable albeit unlikely :)

gsmet:
I just don't see the value in rushing things except having later issues.

We can easily merge as soon as we want it.

geoand:
Yeah, like I said, I understand :)

geoand:
Force pushed in order to use latest ASM

Sanne:
The ORM/bytebuddy upgrade is available now:
 - https://github.com/quarkusio/quarkus/pull/9006

Hopefully we can upgrade ASM soon, as otherwise I can't use the improvements from the latest ORM. (it now *requires* ASM 8)

geoand:
cc @gsmet

PR #40, opened 2020-04-21
BytecodeCreator - add ifValue() to handle single-valued if statements

### LINKED ISSUE #39: Add more if statements
Currently, we only have `ifNonZero` and `ifNull`. It would be useful to add more variants, `ifgt` or `iflt` at least.

mkouba:
- resolves #39

### DIFF
--- a/src/main/java/io/quarkus/gizmo/BytecodeCreator.java
+++ b/src/main/java/io/quarkus/gizmo/BytecodeCreator.java
@@ -487,7 +487,7 @@ public interface BytecodeCreator extends AutoCloseable {
      * @param resultHandle
      * @return The branch result that is used to build the if statement
      */
-    BranchResult ifLowerThanZero(ResultHandle resultHandle);
+    BranchResult ifLessThanZero(ResultHandle resultHandle);
 
     /**
      * An if statement.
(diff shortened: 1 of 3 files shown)


mkouba:
It could be useful to add support for `if_icmpeq` and friends as well...

src/main/java/io/quarkus/gizmo/BytecodeCreator.java
```diff
@@ -559,4 +571,62 @@ default void breakScope() {
      */
     @Override
     default void close() {}
+    
+    /**
+     * The JVM opcode that can be used for {@link BytecodeCreator#ifValue(ResultHandle, IfOpcode)}.
+     * 
+     * @see Opcodes
+     */
+    enum IfOpcode {
```
stuartwdouglas, line 580: I don't really like this, you should not need to know about bytecodes to use Gizmo, I think we should add methods for each of them.

mkouba: Well, you need to know something about bytecode (see for example all those `invokeX()` methods, `ifNonZero()` is used to test booleans, etc.). In fact, you need to know a lot about bytecode in order to produce "effective" classes. I can rework the PR though and add those 6 methods.

mkouba:
I've added few more if statements...

PR #41, opened 2020-05-05
Add a while-loop-statement-like abstraction

mkouba:
I know that it's possible to implement a loop using scopes. However, the API is IMHO not very intuitive and the resulting code is error prone.

I propose to add a simple while loop abstraction.

Compare:
```java
AssignableResultHandle val = method.createVariable(int.class);
method.assign(val, method.load(10));

BytecodeCreator loop = method.createScope();
BranchResult branch = loop.ifNonZero(val);
BytecodeCreator tb = branch.trueBranch();
tb.invokeVirtualMethod(...);
tb.assign(val, tb.invokeStaticMethod(MethodDescriptor.ofMethod(Math.class, "addExact", int.class, int.class, int.class), val, tb.load(-1)));
tb.continueScope(loop);
branch.falseBranch().breakScope(loop);
```
VS
```java
AssignableResultHandle val = method.createVariable(int.class);
method.assign(val, method.load(10));

WhileLoop loop = method.whileLoop(bc -> bc.ifNonZero(val));
BytecodeCreator block = loop.block();
block.invokeVirtualMethod(...);
block.assign(val, block.invokeStaticMethod(MethodDescriptor.ofMethod(Math.class, "addExact", int.class, int.class, int.class), val, block.load(-1)));
```

### DIFF
--- a/src/main/java/io/quarkus/gizmo/WhileLoop.java
+++ b/src/main/java/io/quarkus/gizmo/WhileLoop.java
@@ -1,0 +1,45 @@
+/*
+ * Copyright 2020 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.quarkus.gizmo;
+
+/**
+ * A while loop statement.
+ */
+public interface WhileLoop {
+
+    /**
+     * The block is executed until the condition evaluates to false.
+     * 
+     * @return the while block
+     */
+    BytecodeCreator block();
+
+    /**
+     * The scope returned from this method should be used to continue/break the while statement.
+     * 
+     * <pre>
+     * WhileLoop loop = method.whileLoop(bc -> bc.ifTrue(condition));
+     * BytecodeCreator block = loop.block();
+     * // If counter > 5 we break the loop even if the condition is still true 
+     * block.ifIntegerGreaterThan(counter, block.load(5)).trueBranch().breakScope(loop.scope());
+     * </pre>
+     * 
+     * @return the scope that could be used to skip the current iteration or terminate the statement
+     * @see BytecodeCreator#continueScope(BytecodeCreator)
+     * @see BytecodeCreator#breakScope(BytecodeCreator)
+     */
+    BytecodeCreator scope();
+}
--- a/src/main/java/io/quarkus/gizmo/BytecodeCreator.java
+++ b/src/main/java/io/quarkus/gizmo/BytecodeCreator.java
@@ -17,6 +17,7 @@
 package io.quarkus.gizmo;
 
 import java.util.Objects;
+import java.util.function.Function;
 
 import org.jboss.jandex.FieldInfo;
 import org.jboss.jandex.MethodInfo;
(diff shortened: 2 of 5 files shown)


src/main/java/io/quarkus/gizmo/WhileLoop.java
```diff
@@ -0,0 +1,37 @@
...
+     */
+    BytecodeCreator block();
+
+    /**
+     * 
+     * @return the scope that could be used to skip the current iteration or terminate the statement
+     * @see BytecodeCreator#continueScope(BytecodeCreator)
+     * @see BytecodeCreator#breakScope(BytecodeCreator)
+     */
+    BytecodeCreator scope();
```
stuartwdouglas, line 36: How would you use this? I don't really understand how this bit works. Looking at this impl isn't this just code that is executed after the loop?

mkouba: It's [here](https://github.com/quarkusio/gizmo/pull/41/files#diff-b54ff6ffe788368d2da640abdc1ae2dbR101) and [here](https://github.com/quarkusio/gizmo/pull/41/files#diff-b54ff6ffe788368d2da640abdc1ae2dbR131). You can't use the scope of the `WhileLoop.block()`. Instead, you need to use the scope of the while statement (the same scope is used in the function argument of `WhileLoop`).

Maybe we should add an example to the javadoc...

mkouba: Well, you're actually right that `WhileLoopImpl` is itself the scope we're interested in. However, my idea was to instruct the users to specify the condition and the while block. That's why `WhileLoop` does not extend `BytecodeCreator`... On the other hand, they can always do `WhileLoop.scope().doStuff()` so they can always break things anyway.

stuartwdouglas: Maybe WhileLoop.confinue(BytecodeCreator) and WhileLoop.break(BytecodeCreator) would make more sense?

stuartwdouglas: Well I guess not those exact method names, but you get the idea :-)

mkouba: Yeah, I understand. The problem is that we need to continue/break inside the while block. So the usage is `whileBlock.continueScope(whileLoop.scope())`. Do you get my point?

stuartwdouglas: I mean that the API should look like:

whileLoop.doContinue(whileBlock);

If continue and break are the only valid uses for the ByteCodeCreator then we could just add them to the WhileLoop interface, which avoids exposing the scope and makes it more obvious how it is used.

mkouba: Maybe I'm missing something but I still can't see how this API would work - you need to continue/break inside the while block like in:
```java
while(true) { // -> BranchResult returned from the function passed to BytecodeCreator.whileLoop()
  break; // -> while block, BytecodeCreator returned from WhileLoop.block()
} 
```
And a gizmo equivalent:
```java
WhileLoop loop = method.whileLoop(bc -> bc.ifTrue(bc.load(true)));
loop.block().continueScope(loop.scope());
```

We could return the `BytecodeCreator` for the corresponding while block directly from the `BytecodeCreator.whileLoop()` method but in that case `BytecodeCreator.continueScope()/breakScope()` would not work because the while block is the true branch of the `BranchResult` returned from the function passed to `BytecodeCreator.whileLoop()`.

stuartwdouglas: https://github.com/quarkusio/gizmo/compare/master...stuartwdouglas:while-loop?expand=1 is what I mean

mkouba: Ah, now I understand but the API still looks quite ugly. Anyway, it's better than what we have here. So pls add your commit to this PR.

Also we can't prevent the `BytecodeCreator` "leak" from the condition function. But that's probably an acceptable design flaw ;-).

stuartwdouglas:
I got 'permission denied' when trying to push to your branch. Maybe github is not going to let you just push to other peoples branches anymore

mkouba:
@stuartwdouglas "Allow edits by maintainers" was unchecked for some reason...

PR #68, opened 2020-12-19
add constructor methods to simplified call

### LINKED ISSUE #65: Simplify creating constructors
The syntax is quite arcane and is difficult to get right, in particular when needing to call super constructors.

dufoli:
fix #65

### DIFF
--- a/src/main/java/io/quarkus/gizmo/ClassCreator.java
+++ b/src/main/java/io/quarkus/gizmo/ClassCreator.java
@@ -73,6 +73,14 @@ public class ClassCreator implements AutoCloseable, AnnotatedElement, SignatureE
         this(null, classOutput, name, signature, superClass, 0, interfaces);
     }
 
+    public MethodCreator getConstructor(String... parameters) {
+        return getMethodCreator(MethodDescriptor.ofConstructor(className, parameters));
+    }
+
+    public MethodCreator getSuperConstructor(String... parameters) {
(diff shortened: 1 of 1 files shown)


dufoli:
@gsmet @stuartwdouglas @Sanne  can you review it ?

src/main/java/io/quarkus/gizmo/ClassCreator.java
```diff
@@ -73,6 +73,14 @@ public ClassCreator(ClassOutput classOutput, String name, String signature, Stri
         this(null, classOutput, name, signature, superClass, 0, interfaces);
     }
 
+    public MethodCreator getConstructor(String... parameters) {
+        return getMethodCreator(MethodDescriptor.ofConstructor(className, parameters));
+    }
+
+    public MethodCreator getSuperConstructor(String... parameters) {
```
stuartwdouglas, line 80: This does not seem right, you can't create the superclasses constructor in a subclass.

stuartwdouglas (changes requested):

src/main/java/io/quarkus/gizmo/ClassCreator.java
dufoli: hmm you are right. Indeed I start this one then switch to returnValue patch to implement the call super constructor patch. I think this one was just to delete. I will close it.

PR #69, opened 2020-12-19
start to implement test of returnValue

dufoli:
Missing part:
- functionnal creator

- if method contain branch with if false/ true and each part returnValue.
So maybe best is to implement a visitor of methods to check return instead of a global simple check.
same things for while, for... indeed any statement

- thinking to warning instead of exception but the closest for that seems to be ErrorReporter but it seems not implemented.

### DIFF
--- a/src/main/java/io/quarkus/gizmo/MethodCreatorImpl.java
+++ b/src/main/java/io/quarkus/gizmo/MethodCreatorImpl.java
@@ -41,18 +41,32 @@ class MethodCreatorImpl extends BytecodeCreatorImpl implements MethodCreator {
     private final String declaringClassName;
     private final ClassCreator classCreator;
     private String signature;
+    private Boolean isValidMethod;
 
     MethodCreatorImpl(BytecodeCreatorImpl enclosing, MethodDescriptor methodDescriptor, String declaringClassName, ClassCreator classCreator) {
         super(enclosing, true);
         this.methodDescriptor = methodDescriptor;
         this.declaringClassName = declaringClassName;
         this.classCreator = classCreator;
+        this.isValidMethod = methodDescriptor.getReturnType().equals("V");
     }
 
     @Override
     public MethodCreator addException(String exception) {
         exceptions.add(exception.replace('.', '/'));
         return this;
+    }
+
+    @Override
+    public void returnValue(ResultHandle returnValue) {
+        super.returnValue(returnValue);
+        isValidMethod = true;
+    }
+
+    @Override
+    public void throwException(ResultHandle exception) {
+        super.throwException(exception);
+        isValidMethod = true;
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


dufoli:
Hello, code is ready with PR. I have just a doubt about impact of the classCreator close.

https://github.com/quarkusio/gizmo/blob/bda4d0794f8b808c095c1351ba6bc48f92f50403/src/main/java/io/quarkus/gizmo/BytecodeCreatorImpl.java#L987
Else it is mostly store and load return handle so no impact because it call visitor ...

dufoli:
I ma thinking to split isValid to isMissingReturnValue and isMissingSuper or this call in ctor.
To auto add call to object one by default:
```
            ctor.invokeSpecialMethod(
                    MethodDescriptor.ofConstructor("java.lang.Object"),
                    ctor.getThis());
```
wdyt ?

src/main/java/io/quarkus/gizmo/MethodCheckerVisitor.java
```diff
@@ -0,0 +1,220 @@
...
+        }
+    }
+
+    @Override
+    public void visitParameter(String name, int access) {
+    }
+
+    @Override
+    public AnnotationVisitor visitAnnotationDefault() {
+        return new DummyVisitor();
```
Dufgui, line 55: why not return null like https://github.com/llbit/ow2-asm/blob/master/src/org/objectweb/asm/MethodVisitor.java ?

src/main/java/io/quarkus/gizmo/MethodCheckerVisitor.java
```diff
@@ -0,0 +1,220 @@
...
+    public void visitTypeInsn(int opcode, String type) {
+    }
+
+    @Override
+    public void visitFieldInsn(int opcode, String owner, String name, String descriptor) {
+    }
+
+    @Override
+    public void visitMethodInsn(int opcode, String owner, String name, String descriptor) {
+    }
```
Dufgui, line 107: you don't need isSuperOrThisCalled = true; like other visitMethodInsn ?

dufoli: ok I just added it.

src/main/java/io/quarkus/gizmo/MethodCheckerVisitor.java
dufoli: current code do not handle it :
AnnotationUtils.visitAnnotationValue do not check if first param is null.
But you are right asm standard said that it can return null if no visitor is needed.
Other solution is to return null and check it in
https://github.com/quarkusio/gizmo/blob/bda4d0794f8b808c095c1351ba6bc48f92f50403/src/main/java/io/quarkus/gizmo/AnnotationUtils.java#L9

dufoli: @gsmet @stuartwdouglas  I provide 2 solutions for that. current code with dummy visitor or return null and handle it in AnnotationUtils. What do you prefer ?
Anyway, I need a full review before going further.

dufoli:
@gsmet @stuartwdouglas @Sanne  can you review it ?

src/main/java/io/quarkus/gizmo/MethodCheckerVisitor.java
```diff
@@ -0,0 +1,223 @@
...
+
+import org.objectweb.asm.AnnotationVisitor;
+import org.objectweb.asm.Attribute;
+import org.objectweb.asm.Handle;
+import org.objectweb.asm.Label;
+import org.objectweb.asm.MethodVisitor;
+import org.objectweb.asm.Opcodes;
+import org.objectweb.asm.TypePath;
+
+public class MethodCheckerVisitor extends MethodVisitor {
```
stuartwdouglas, line 11: Why is this done as a visitor, rather than just checking that the bytecode is valid when it is generated?

dufoli: imagine you have a code with 
```
public int myfunct(boolean foo, boolean bar) {
 if (foo) return 1
else if (bar) return 0;
else
throw new RuntimeException();
}
```
so you have 2 branchs of return to check indeed.
My first think was to check that last operation was a return and it failed because of that.
You have the exception case too...

dufoli: you can not check just the methodMody, you have to enter in operation to check them.
Or you mean that I check generated code instead of AST ?

PR #81, opened 2021-06-15
Complete annotation creation

Ladicek:
The code in `AnnotationUtils` is completed to properly handle
all possible Jandex `AnnotationValue` instances, including
`AnnotationInstance` values (nested annotations).

With this, `AnnotationCreator` can store all `AnnotationValue`
and `AnnotationInstance` values, which is used to simplify
`AnnotatedElement.addAnnotation(AnnotationInstance)`.


Ladicek:
Motivation is https://github.com/smallrye/smallrye-fault-tolerance/issues/440#issuecomment-860733991

stuartwdouglas (approved):

PR #92, opened 2021-11-08
Load classes from the TCCL

stuartwdouglas:


stuartwdouglas:
This has been a constant issue, rather than keeping on fixing it piecemeal in Quarkus IMHO we should just fix it here.

src/main/java/io/quarkus/gizmo/BytecodeCreator.java
```diff
@@ -267,21 +269,68 @@ default ResultHandle load(Enum<?> val) {
...
      * Returns a {@link ResultHandle} representing the specified class
      *
+     * This will be loaded from the TCCL, rather than via bytecode reference.
+     *
      * @param val The class to load
      * @return A {@link ResultHandle} representing the specified class
      */
     default ResultHandle loadClass(Class<?> val) {
-        return loadClass(val.getName());
+        return loadClass(val.getName(),true);
```
mkouba, line 278: I'm not entirely sure the TCCL should be used by default. It's probably OK in most cases but anyway is ok for all cases? I'm talking about backward compatibility.

src/main/java/io/quarkus/gizmo/BytecodeCreatorImpl.java
```diff
@@ -342,7 +354,16 @@ public ResultHandle loadClass(String className) {
...
         }
         if (primitiveType == null) {
-            return new ResultHandle("Ljava/lang/Class;", this, Type.getObjectType(className.replace('.', '/')));
+            if (useTccl) {
+                if (cachedTccl == null) {
+                    ResultHandle currentThread = invokeStaticMethod(THREAD_CURRENT_THREAD);
+                   cachedTccl = invokeVirtualMethod(THREAD_GET_TCCL, currentThread);
+                }
+                return invokeStaticMethod(CL_FOR_NAME,
+                                load(className), load(false), cachedTccl);
```
mkouba, line 363: Shouldn't we allow users to speficy the `initialize` value?

mkouba:
> This has been a constant issue, rather than keeping on fixing it piecemeal in Quarkus IMHO we should just fix it here.

+10 to fix the problem in gizmo.

src/main/java/io/quarkus/gizmo/BytecodeCreatorImpl.java
stuartwdouglas: Not really, its kinda pointless IMHO unless you really want to force static init to run. If they really want to do that they can call the class manually.

src/main/java/io/quarkus/gizmo/BytecodeCreator.java
stuartwdouglas: If there are cases in Quarkus where this causes problems we can change them to use the explicit version, however I don't think this will be the case. It's much more likely that this change will fix bugs we have not discovered yet.

mkouba: It might fix some issues but it just does not feel OK to me. What was previously a simple `ldc` from the constant pool would become `Thread.currentThread().getContextClassLoader()` + `Class.forName()` (with TCCL cached in the current scope IIUIC) and there's no way to detect this change (except if your code breaks). But maybe I'm just "hypersensitive" ;-)

@geoand @gsmet @Ladicek @Sanne and others WDYT?

src/main/java/io/quarkus/gizmo/BytecodeCreatorImpl.java
mkouba: My point is that `java.lang.Class.forName(String)` has `initialize=true` so it's IMO kind of expected to force static init.

src/main/java/io/quarkus/gizmo/BytecodeCreator.java
geoand: I love the idea of fixing the problem once and for all, but I share Martin's concern here, that `loadClass` should be `ldc` by default.

Sanne: Isn't it (generally) expected that all code gets load from the contextual classloader? That's how a normal Java program works.

I think this patch makes sense as a default; I also share some nervousness about changing such a core detail but let's see CI: failures might indeed indicate some need for some code to not use the default and we can address that, this still makes sense conceptually.

Ladicek: It is not how normal Java programs work. In a normal situation, when class A refers to class B (through normal linguistic means, not through reflection), then B is loaded by the same classloader that loaded A. TCCL doesn't even have to be set (and `Class.forName()` apparently uses the bootstrap CL when passed `null`).

Now, it is rather common to prefer loading classes from TCCL in applications that are "containers" (be it application servers, or Quarkus). If Gizmo serves Quarkus and doesn't strive to be generic, then this change probably makes sense. I do have some mixed feelings about it too, but can't really articulate those.

geoand: > It is not how normal Java programs work. In a normal situation, when class A refers to class B (through normal linguistic means, not through reflection), then B is loaded by the same classloader that loaded A. TCCL doesn't even have to be set (and Class.forName() apparently uses the bootstrap CL when passed null).

+1

Sanne: doh right. Sorry - best not try to rush emails before coffee :)

still, what I said is generally the expectation in Quarkus specifically, as the contect classloader is _the_ runtime classloader and delegating to it should be the safe choice, as it then is able to forward to the right classloaders, and take into account dev/prod differences.

stuartwdouglas: I will be incredibly surprised if this causes any issues in Quarkus, and if it does it means there is something very wrong going on that should be fixed anyway. 

Given that people seem to be worried about changing this I could just add a new method and change every call in Quarkus over to use the new one, and blacklist the current method.

gsmet:
So what are we doing with this? If we want to get it in, I would rather get it in earlier rather than later so that we have ample time to adjust Quarkus and get everything stabilized for 2.6.0.CR1.

@mkouba could you have a look at Stuart's latest commit?

Ladicek:
I didn't notice the change, but the current state of this PR looks good to me.

mkouba:
> So what are we doing with this? If we want to get it in, I would rather get it in earlier rather than later so that we have ample time to adjust Quarkus and get everything stabilized for 2.6.0.CR1.
> 
> @mkouba could you have a look at Stuart's latest commit?

Looks good now. In fact, this PR itself will have zero effect because the `loadClass()` method behaves like before. It will be the follow-up PR in quarkus that might break things.

mkouba (approved):

src/test/java/io/quarkus/gizmo/LoadClassTestCase.java
```diff
@@ -36,6 +36,19 @@ public void testLoadClass() throws Exception {
         Assert.assertEquals(String.class, myInterface.get());
     }
 
+    @Test
+    public void testLoadClassFromTCCL() throws Exception {
```
mkouba, line 40: Actually, shouldn't this test set the TCCL?

mkouba: I mean can we be sure that the TCCL is always set when this test is executed?

PR #93, opened 2021-11-25
Add API to emit invokespecial with an InterfaceMethodref

Ladicek:


Ladicek:
@stuartwdouglas One thing I wanted to ask: I noticed that `FunctionCreatorImpl.FunctionBytecodeCreator` overrides `invokeSpecialMethod` to handle some special case. I'm not sure if it needs to override `invokeSpecialInterfaceMethod` in the same way. I think not, but WDYT?

stuartwdouglas (commented):
With the current should do the same thing for the function creator. Basically this allows invokeSpecial to be used from the function.

Really though I should just re-write the function support to use method handles...

stuartwdouglas:
Actually even if I do re-write it to use method handles it will still need to be overridden.

Ladicek:
OK, thanks, I'll take a look.

Ladicek:
Took me a little while to understand what's the purpose of these overrides, but it made sense in the end :-)

Added an override of `invokeSpecialInterfaceMethod` that works similarly to the override of `invokeSpecialMethod`, except for superinterfaces instead of the superclass. I adjusted the accessor function generator to support both, as the code is mostly identical, it's just about emitting the correct `invokespecial` format.

stuartwdouglas (approved):
LGTM. I have started to look at doing the function stuff as method references, however it turns out it is more complex than I thought it would be.

Ladicek:
Emitting pseudonymous classes may be old-school, but works well :-)

If you could do a release (maybe together with the `loadFromTCCL` PR) in the next few days, that would be nice. (I need it for https://github.com/quarkusio/quarkus/pull/21692.)

stuartwdouglas:
I don't have release perms, we need @gsmet @geoand or @cescoffier to do the release

geoand:
I pressume we want to release `1.0.10.Final`?

gsmet:
Just checking, we don't want https://github.com/quarkusio/gizmo/pull/92 first?

stuartwdouglas:
I have merged it, we should be good to go now.

geoand:
@gsmet do you want to release or should I? `1.0.10.Final` right?

Ladicek:
To the best of my knowledge, it should be 1.0.10.Final indeed.

gsmet:
I'll do it.

PR #100, opened 2022-03-10
Make FieldCreator support signature

rsynek:
Imagine the following field, which injects a smallrye reactive messaging `Emitter`:
```
@Channel(value = "event")
Emitter<Event> eventEmitter;
```
Smallrye requires the `Emitter` to have a type argument that represents the message the `Emitter` is supposed to send.

Such a field cannot be generated without the `setSignature()` call, which is currently unavailable in the `FieldCreator` interface:
```
FieldCreator fieldCreator = classCreator.getFieldCreator("eventEmitter", Emitter.class);
fieldCreator.setSignature("Lorg/eclipse/microprofile/reactive/messaging/Emitter<Lorg/something/messaging/Event;>;");
```

### DIFF
--- a/src/main/java/io/quarkus/gizmo/FieldCreator.java
+++ b/src/main/java/io/quarkus/gizmo/FieldCreator.java
@@ -17,7 +17,7 @@
 package io.quarkus.gizmo;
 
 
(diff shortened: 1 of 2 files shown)


rsynek:
@mkouba or @Ladicek could you kindly have a look?

mkouba (approved):

PR #107, opened 2022-04-29
Allow using Maps to specify nested annotations

Christopher-Chianelli:
All Java Annotations extend java.lang.annotation.Annotation,
which defines a method `annotationType` that return the type
of the annotation (and not the proxy class). As such, it can
be used as a key in a Map to specify the type of the Annotation
to create, with the remaining keys being the elements of the
Annotation.


Ladicek:
I personally think that adding a method `AnnotationCreator addNested(String annotationType)` (that returns a new `AnnotationCreator` for the nested annotation) would be nicer, but this works too :-)

PR #108, opened 2022-05-02
Add nested AnnotationCreator alternative; add Javadoc for AnnotationCreator

Christopher-Chianelli:
Add a new method to support creating nested annotations using
AnnotationCreators. It works by creating a new AnnotationCreator
and adding its value map as a value to the parent AnnotationCreator
(thus any changes made to the nested AnnotationCreator update the
map in the parent's values). Also added Javadoc so these features
are discoverable.

### DIFF
--- a/src/main/java/io/quarkus/gizmo/AnnotationCreator.java
+++ b/src/main/java/io/quarkus/gizmo/AnnotationCreator.java
@@ -16,8 +16,65 @@
 
 package io.quarkus.gizmo;
 
+import java.lang.annotation.Annotation;
+import java.util.Map;
+import java.util.function.Consumer;
+
 public interface AnnotationCreator {
 
+    /**
+     * Add a new element with the given {@code name} and {@code value}. If {@code value} is a {@link Map},
+     * it is treated as a nested annotation, with the map's entries being the nested annotation's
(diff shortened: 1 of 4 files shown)


Ladicek:
Hi @Christopher-Chianelli, I looked into this too, because I figured my suggestion to add `AnnotationCreator addNested` was not very well thought through. Instead, I came up with this: https://github.com/quarkusio/gizmo/compare/main...Ladicek:nested-annotation-creator?expand=1 I believe it nicely replaces the `Map` usage as well as the various `addNested` methods. WDYT?

Christopher-Chianelli:
Greetings @Ladicek ; Looking at https://github.com/quarkusio/gizmo/compare/main...Ladicek:nested-annotation-creator?expand=1 , I agree it nicely replaces the Map usage and offer a better API than this PR. `Map` was used since I needed a type that:

- Users are already familiar with
- Supports keys and values
- Is easy to create (Jandex `AnnotationInstance` is not easy to create)

`AnnotationCreator` also support each of the above, while also naturally representing annotations. The code in the branch LGTM.

Ladicek:
I'll submit a PR then, thank you!

mkouba:
I'm going to close this PR in favor of https://github.com/quarkusio/gizmo/pull/111.

PR #111, opened 2022-05-05
Allow using AnnotationCreator to represent nested annotations

Ladicek:


Ladicek:
This replaces the previous change (#107) and is an alternative to another PR (#108). See #108 for the original author of #107 expressing an agreement with this.

src/main/java/io/quarkus/gizmo/AnnotationUtils.java
```diff
@@ -28,24 +26,15 @@ static void visitAnnotationValue(AnnotationVisitor visitor, String key, Object v
                 visitAnnotationValue(nestedVisitor, annotationValue.name(), annotationValue);
             }
             nestedVisitor.visitEnd();
-        } else if (value instanceof Map) {
```
mkouba, line 31: Hm, so this change is not bacward compatible, right?

mkouba (approved):
Looks good.

PR #113, opened 2022-05-06
Add method to set parameter names

### LINKED ISSUE #112: Support setting method parameter names
I need to generate methods with Bean Validation annotations, such as `foo(@NotEmpty String param)` and if I can't set the method parameter name, then I can't obtain the right validation constraint cause (`param`).

Ladicek:
Fixes #112

### DIFF
--- a/src/main/java/io/quarkus/gizmo/MethodCreator.java
+++ b/src/main/java/io/quarkus/gizmo/MethodCreator.java
@@ -57,4 +57,19 @@ public interface MethodCreator extends MemberCreator<MethodCreator>, BytecodeCre
 
     AnnotatedElement getParameterAnnotations(int param);
 
+    /**
+     * Sets names of all this method's parameters. The length of {@code parameterNames}
+     * must be equal to the number of this method's parameters. Removes previously set
+     * parameter names if {@code parameterNames} is {@code null}.
+     * <p>
+     * It is required to set all parameter names in one go, because parameter names are stored in bytecode
+     * as a simple sequence, without positions.
+     * <p>
+     * When generating a method with mandated or synthetic parameters (such as an inner
+     * class constructor), remember that these parameters must have names too.
+     *
+     * @param parameterNames names of all parameters of this method
+     */
+    void setParameterNames(String[] parameterNames);
+
 }
--- a/src/main/java/io/quarkus/gizmo/MethodCreatorImpl.java
+++ b/src/main/java/io/quarkus/gizmo/MethodCreatorImpl.java
@@ -19,6 +19,7 @@ package io.quarkus.gizmo;
 import java.lang.annotation.RetentionPolicy;
 import java.lang.reflect.Modifier;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.LinkedHashMap;
 import java.util.List;
@@ -41,6 +42,8 @@ class MethodCreatorImpl extends BytecodeCreatorImpl implements MethodCreator {
     private final String declaringClassName;
     private final ClassCreator classCreator;
     private String signature;
+
+    private String[] parameterNames;
 
     MethodCreatorImpl(BytecodeCreatorImpl enclosing, MethodDescriptor methodDescriptor, String declaringClassName, ClassCreator classCreator) {
         super(enclosing, true);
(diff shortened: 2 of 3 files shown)


src/main/java/io/quarkus/gizmo/MethodCreatorImpl.java
```diff
@@ -75,6 +78,22 @@ public AnnotatedElement getParameterAnnotations(int param) {
         return p;
     }
 
+    @Override
+    public void setParameterNames(String[] parameterNames) {
+        if (parameterNames == null) {
```
gastaldi, line 83: Shouldn't it set `this.parameterNames` to `null`?

src/main/java/io/quarkus/gizmo/MethodCreator.java
```diff
@@ -57,4 +57,15 @@ default MethodCreator addException(Class<?> exception) {
 
     AnnotatedElement getParameterAnnotations(int param);
 
+    /**
+     * Sets names of all this method's parameters. The length of {@code parameterNames} must be equal
+     * to the number of this method's parameters. Does nothing if {@code parameterNames} is {@code null}.
```
dmlloyd, line 62: There are cases where the number of parameter names can or should be less than the number of parameters, e.g. inner class constructors. Not sure if that's too nitpicky but the constraint isn't completely correct.

src/main/java/io/quarkus/gizmo/MethodCreatorImpl.java
Ladicek: That's a good idea. Removing what was already set may occasionally be useful.

src/main/java/io/quarkus/gizmo/MethodCreatorImpl.java
```diff
@@ -75,6 +78,22 @@ public AnnotatedElement getParameterAnnotations(int param) {
...
+            return;
+        }
+
+        if (methodDescriptor.getParameterTypes().length != parameterNames.length) {
```
dmlloyd, line 87: Should probably be `<` not `!=` given above comment.

src/main/java/io/quarkus/gizmo/MethodCreatorImpl.java
Ladicek: Done.

src/main/java/io/quarkus/gizmo/MethodCreator.java
Ladicek: Synthetic and mandated parameters also have names, otherwise you can't match the name with the correct parameter. The enclosing instance parameter in case of inner class constructors are typically called `this$N` (where N >= 0).

Ladicek: Technically what we could allow is omitting names for parameters at the _end_ of the parameter list. But that's hard to express without losing this check.

Or we could allow setting parameter names by position and synthesize names that were not set...

Ladicek: I just realized I was wrong above. Synthetic and mandated parameters only have names in the local variable table. In the Parameters attribute, they do not.

I could perhaps check the number of parameter names against the signature, if present, but if not, I don't think this can meaningfully be checked. I guess I'll just have to remove this check for good and perhaps add some documentation.

Ladicek: Added a little bit of documentation.

FroMage (approved):
LGTM.

src/main/java/io/quarkus/gizmo/MethodCreator.java
dmlloyd: I reviewed the specification and I revise my assessment. It says:

> The value of the parameters_count item indicates the number of parameter descriptors in the method descriptor ([§4.3.3](https://docs.oracle.com/javase/specs/jvms/se17/html/jvms-4.html#jvms-4.3.3)) referenced by the descriptor_index of the attribute's enclosing method_info structure.
>
> _This is not a constraint which a Java Virtual Machine implementation must enforce during format checking ([§4.8](https://docs.oracle.com/javase/specs/jvms/se17/html/jvms-4.html#jvms-4.8)). The task of matching parameter descriptors in a method descriptor against the items in the parameters array below is done by the reflection libraries of the Java SE Platform._

So I agree, the PR is OK as-is.

dmlloyd (approved):

src/main/java/io/quarkus/gizmo/MethodCreatorImpl.java
```diff
@@ -119,6 +139,11 @@ public void write(ClassVisitor file) {
                 av.visitEnd();
             }
         }
+        if (parameterNames != null) {
+            for (String parameterName : parameterNames) {
+                visitor.visitParameter(parameterName, 0);
```
mkouba, line 144: @Ladicek The javadoc of `MethodVisitor.visitParameter()` states that `only {@code ACC_FINAL}, {@code ACC_SYNTHETIC}    or/and {@code ACC_MANDATED} are allowed`. What does the zero represent?

Ladicek: Those are bit flags, so `0` means no flag set. As in, not final, not synthetic and not mandated.

Issue #66, opened 2020-12-18
Simplify the handling of parameterized types


metacosm:
Right now, you need to use `signature` passing it a complex string describing how your type is parameterized. It would be nice if that complexity could be hidden away.

dufoli:
currently, when Class is parameterized, we need to add signature like that:

```
String signature = String.format("Ljavax/enterprise/util/AnnotationLiteral<L%1$s;>;L%1$s;",
        eventAnnotationLiteral.getName().toString().replace('.', '/'));

ClassCreator literalClassCreator = ClassCreator.builder().classOutput(classOutput)
        .className(literalClassName)
        .signature(signature)
        .superClass(AnnotationLiteral.class)
        .interfaces(eventAnnotationLiteral.getName().toString())
        .build();
```

I have introduce on PR a signature generation.

dufoli:
ok so now code is ready for review. I do not think it is ready for merge but size of patch need review to know If I go on right direction. I try to not break API which lead to some stuff
- formalType is added after MethodDescriptor to avoid break of signature of current API
- I discover that classCreator use slashed type and methodCreator used jvm types (thanks to methodDescriptor). I do not touch it because it have impact on public API but I translate each time I need. I was thinking to introduce a class Descriptor... for consistency
- I put deprecated on signature setter 
- I introduce a pseudo format to add generics in method return type. It is just an idea (not very happy of it currently) idea is to have "List<String>" which produce List return for descriptor and keep generics. I am thinking to remove it totally. It was during my early code. But prefer to know if I can work to improve that and keep it and just throw it.

basic API is
```
//public <K> String transform(K thing) { return K.toString());
// signature must be <K:Ljava/lang/Object;>(TK;)Ljava/lang/String;"
try (ClassCreator creator = ClassCreator.builder().classOutput(cl).className("com.MyTest").formalType("K").build()) {
    MethodCreator method = creator.getMethodCreator("transform", "java.lang.String", "K").formalType("K");
    ResultHandle message = method.invokeStaticMethod(MethodDescriptor.ofMethod(Object.class.getName(), "toString", "Ljava/lang/String;"));
    method.returnValue(message);
}
```
I was thinking to a system to forward formalType from creator to method but it is not that simple. because some methods needs it and other don't. But Idea was to add it to classCreator and auto add to method if param used it but for that I must be able to go to classCreator from methodCreator if I have param with one letter and match with class one.
Anyway, can be a feature for futur.

[...]

PR #116, opened 2022-06-15
Support Parameterized Types in Methods

Sgitario:
This pull request handles the support of parameterized types when creating methods. Example:
```java
public List<String> methodName() {
}
```

Where `List<String>` is a parameterized type. 

In order to do this, we need to amend the `signature` of the method as the descriptor does not support it (it says class not found when trying to declare `Ljava.util.List<Ljava.lang.String;>;`). 

In order to resolve this, we have added a wrapper type `ParameterizedClass` where we can specify the wrapper type and the param types. Usage:

```
MethodCreator method = creator.getMethodCreator("methodName", new ParameterizedClass(List.class, String.class));
```

This pull request partially resolves: https://github.com/quarkusio/gizmo/issues/66 (which is intended to be for classes too)
Moreover, these changes are inspired by the changes in https://github.com/quarkusio/quarkus/pull/26074 (that there will be deprecated once gizmo dependency is bumped).

I've verified these changes in Quarkus upstream too (reverting the changes of the mentioned Quarkus pull request) and it worked fine.

### DIFF
--- a/src/main/java/io/quarkus/gizmo/DescriptorUtils.java
+++ b/src/main/java/io/quarkus/gizmo/DescriptorUtils.java
@@ -38,6 +38,16 @@ public class DescriptorUtils {
             void.class
     };
 
+    public static String methodTypesToSignature(Object returnType, Object[] params) {
+        StringBuilder sb = new StringBuilder("(");
+        for (Object i : params) {
+            sb.append(objectToSignature(i));
+        }
+        sb.append(")");
+        sb.append(objectToSignature(returnType));
+        return sb.toString();
+    }
+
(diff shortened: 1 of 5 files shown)


mkouba:
So the `MethodCreator` implements the `SignatureElement` and allows you to set the method signature. I think that it would be more convenient to provide a signature util method that would accept jandex `org.jboss.jandex.Type` params.

Also it seems that `ParameterizedClass` does not support nested params, or?

mkouba:
FTR I'm working on a pull request that introduces a `SignatureBuilder` API (inspired by ideas/work from @Ladicek).

Sgitario:
Ok, closing.

PR #119, opened 2022-08-26
Upgrade to Jandex 3.0.0

Ladicek:
This is technically a breaking change, so would require at least a 1.2 release. Also, it isn't really necessary, Gizmo doesn't use anything from Jandex that changed in a breaking manner, so Gizmo compiled against Jandex 2.4 works with Jandex 3.0 at runtime just fine. I figured I'll submit a PR anyway, as we'll want everyone to move to Jandex 3.0 over time :-)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -35,7 +35,7 @@
 
     <properties>
         <version.asm>9.3</version.asm>
-        <version.jandex>2.4.3.Final</version.jandex>
+        <version.jandex>3.0.0</version.jandex>
         <version.junit>4.13.2</version.junit>
         <nexus-staging-maven-plugin.version>1.6.13</nexus-staging-maven-plugin.version>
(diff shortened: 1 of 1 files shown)


gsmet (approved):

gsmet:
@Ladicek do you want me to release a 1.2.0?

Ladicek:
Not needed, but won't hurt either :-)

gsmet:
Done (it should hit Central in ~20mn). That will be less exclusions to add.

BTW, when upgrading in the core Quarkus repo, let's add the old artifact to the banned dependencies so that we make sure it doesn't end up in our dependencies.

Ladicek:
I already have all the necessary exclusions set up (and Enforcer configured), but indeed one less exclusion is always good. Thanks!

PR #123, opened 2022-10-07
Introduce BytecodeCreator#ifThenElse()

mkouba:
The `IfThenElse` construct aims to provide a less error-prone way of generating `if-then-else` statements.

Currently, if you want to use gizmo to generate something like:
```java
String val;
if (val.equals("foo")) {
  val = "FOO";
} else if (val.equals("bar")) {
  val = "BAR!";
} else if (val.equals("baz")) {
   var = "BAZ!";
} else {
  val = "OTHER!";
}
```
...then the gizmo code looks like:
```java
BranchResult ifFoo = method.ifTrue(Gizmo.equals(method, method.getMethodParam(0), method.load("foo")));
BytecodeCreator ifFooTrue = ifFoo.trueBranch();
ifFooTrue.assign(ret, ifFooTrue.load("FOO!"));

BranchResult ifBar = ifFoo.falseBranch().ifTrue(Gizmo.equals(ifFoo.falseBranch(), method.getMethodParam(0), ifFoo.falseBranch().load("bar")));
BytecodeCreator ifBarTrue = ifBar.trueBranch();
ifBarTrue.assign(ret, ifBarTrue.load("BAR!"));

BranchResult ifBaz = ifBar.falseBranch().ifTrue(Gizmo.equals(ifBar.falseBranch(), method.getMethodParam(0), ifBar.falseBranch().load("baz")));
BytecodeCreator ifBazTrue = ifBaz.trueBranch();
ifBazTrue.assign(ret, ifBazTrue.load("BAZ!"));

BytecodeCreator elseBranch = ifBaz.falseBranch();
elseBranch.assign(ret, elseBranch.load("OTHER!"));
```
Note that you need to be extra careful which `BytecodeCreator` is used when, i.e. notice how the `BranchResult ifBaz` is created.

And the version with `IfThenElse`:
```java
IfThenElse ifValue = method.ifThenElse(Gizmo.equals(method, method.getMethodParam(0), method.load("foo")));

BytecodeCreator ifFooNext = ifValue.then();
ifFooNext.assign(ret, ifFooNext.load("FOO!"));

BytecodeCreator ifBar = ifValue.elseIf(b -> Gizmo.equals(b, method.getMethodParam(0), b.load("bar")));
ifBar.assign(ret, ifBar.load("BAR!"));

BytecodeCreator ifBaz = ifValue.elseIf(b -> Gizmo.equals(b, method.getMethodParam(0), b.load("baz")));
ifBaz.assign(ret, ifBaz.load("BAZ!"));

BytecodeCreator elseThen = ifValue.elseThen();
elseThen.assign(ret, elseThen.load("OTHER!"));
```

### DIFF
--- a/src/main/java/io/quarkus/gizmo/BytecodeCreator.java
+++ b/src/main/java/io/quarkus/gizmo/BytecodeCreator.java
@@ -40,8 +40,8 @@ public interface BytecodeCreator extends AutoCloseable {
      * Invokes a virtual method, and returns a {@link ResultHandle} with the result, or null if the method is void.
      *
      * @param descriptor The method descriptor
-     * @param object     A {@link ResultHandle} representing the object to invoke on
-     * @param args       The method parameters
+     * @param object A {@link ResultHandle} representing the object to invoke on
+     * @param args The method parameters
      * @return The method result, or null if a void method
      */
     ResultHandle invokeVirtualMethod(MethodDescriptor descriptor, ResultHandle object, ResultHandle... args);
@@ -50,8 +50,8 @@ public interface BytecodeCreator extends AutoCloseable {
      * Invokes a virtual method, and returns a {@link ResultHandle} with the result, or null if the method is void.
      *
      * @param descriptor The method descriptor
-     * @param object     A {@link ResultHandle} representing the object to invoke on
-     * @param args       The method parameters
+     * @param object A {@link ResultHandle} representing the object to invoke on
+     * @param args The method parameters
      * @return The method result, or null if a void method
      */
     default ResultHandle invokeVirtualMethod(MethodInfo descriptor, ResultHandle object, ResultHandle... args) {
@@ -62,8 +62,8 @@ public interface BytecodeCreator extends AutoCloseable {
      * Invokes a interface method, and returns a {@link ResultHandle} with the result, or null if the method is void.
      *
      * @param descriptor The method descriptor
-     * @param object     A {@link ResultHandle} representing the object to invoke on
-     * @param args       The method parameters
+     * @param object A {@link ResultHandle} representing the object to invoke on
+     * @param args The method parameters
      * @return The method result, or null if a void method
      */
     ResultHandle invokeInterfaceMethod(MethodDescriptor descriptor, ResultHandle object, ResultHandle... args);
@@ -72,8 +72,8 @@ public interface BytecodeCreator extends AutoCloseable {
      * Invokes a interface method, and returns a {@link ResultHandle} with the result, or null if the method is void.
      *
      * @param descriptor The method descriptor
-     * @param object     A {@link ResultHandle} representing the object to invoke on
-     * @param args       The method parameters
+     * @param object A {@link ResultHandle} representing the object to invoke on
+     * @param args The method parameters
      * @return The method result, or null if a void method
      */
     default ResultHandle invokeInterfaceMethod(MethodInfo descriptor, ResultHandle object, ResultHandle... args) {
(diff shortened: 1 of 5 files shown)


stuartwdouglas (approved):

Ladicek (approved):
LGTM.

Since `IfThenElseImpl` guards against wrong method call order, it might make sense to enforce calling `then()` before `elseIf()` / `elseThen()` is called?

(Also, some methods are placed a little weird -- the `elseIf(Function)` method in `IfThenElseImpl`, the `ifThenElse()` method in `BytecodeCreator` --, but that's just nitpicking.)

mkouba:
> Since `IfThenElseImpl` guards against wrong method call order, it might make sense to enforce calling `then()` before `elseIf()` / `elseThen()` is called?

That's a good point.

> Also, some methods are placed a little weird -- the elseIf(Function) method in IfThenElseImpl, the ifThenElse() method in BytecodeCreator --, but that's just nitpicking.

Like placing `elseIf(Function)` right after `elseIf(ResultHandle)`? That makes sense. I'm not sure the `ifThenElse()` though. It's right after the simple `if` statements so...?

mkouba:
> > Also, some methods are placed a little weird -- the elseIf(Function) method in IfThenElseImpl, the ifThenElse() method in BytecodeCreator --, but that's just nitpicking.
> 
> Like placing `elseIf(Function)` right after `elseIf(ResultHandle)`? That makes sense. I'm not sure the `ifThenElse()` though. It's right after the simple `if` statements so...?

@Ladicek `elseIf(Function)` actually is right after `elseIf(ResultHandle)` so I'm confused now.

Ladicek:
> > Since `IfThenElseImpl` guards against wrong method call order, it might make sense to enforce calling `then()` before `elseIf()` / `elseThen()` is called?
> 
> That's a good point.

On the other hand, sometimes you may want to do just `elseThen()` to avoid having to negate the condition expression. So maybe just ignore this :-)

> Like placing `elseIf(Function)` right after `elseIf(ResultHandle)`? 

Yeah.

> I'm not sure the `ifThenElse()` though. It's right after the simple `if` statements so...?

It's probably because the methods are already a bit misordered, because these methods are _below_ the `ifThenElse()` method:

- `instanceOf`
- `instanceOf`
- `ifIntegerLessEqual`
- `ifReferencesEqual`

As I said, it's just nitpicking, feel free to ignore that.

Ladicek:
> @Ladicek `elseIf(Function)` actually is right after `elseIf(ResultHandle)` so I'm confused now.

In the interface, yes, but not in the implementation class.

mkouba:
> On the other hand, sometimes you may want to do just elseThen() to avoid having to negate the condition expression. So maybe just ignore this :-)

Hm, I think it's actually reasonable because using only `elseThen()` is confusing and harder to read. And it does not make sense to use the `IfThenElse` for simple stuff... I mean you can just use `ifFalse()` instead.

> In the interface, yes, but not in the implementation class.

Ah, you're really strict today!

Ladicek:
> > On the other hand, sometimes you may want to do just elseThen() to avoid having to negate the condition expression. So maybe just ignore this :-)
> 
> Hm, I think it's actually reasonable because using only `elseThen()` is confusing and harder to read. And it does not make sense to use the `IfThenElse` for simple stuff... I mean you can just use `ifFalse()` instead.

Good point.

> > In the interface, yes, but not in the implementation class.
> 
> Ah, you're really strict today!

:laughing:

Ladicek (approved):

Issue #124, opened 2022-10-07
Document basic usage of Gizmo API


mkouba:
Add some basic API usage docs in the README.

PR #125, opened 2022-10-07
Add a brief "basic usage" document

mkouba:
This is the first attempt to document the basic usage of Gizmo API and resolve https://github.com/quarkusio/gizmo/issues/124.

### DIFF
--- a/USAGE.adoc
+++ b/USAGE.adoc
@@ -1,0 +1,257 @@
+= Basic Usage
+
+== Creating a Class
+
+In order to generate bytecode of a new class we need to create a `ClassCreator` instance.
+We can start with the convenient `ClassCreator.Builder`.
+
+[source,java]
+----
+ClassCreator creator = ClassCreator.builder() <1>
+                             .classOutput(cl) <2>
+                             .className("com.MyTest") <3>
+                             .build() <4>
+----
+<1> Create a new `ClassCreator.Builder` instance.
+<2> Set the `ClassOutput` that is used to write the bytecode of the class.
+<3> Specify the fully-qualified class name.
+<4> Build the `ClassCreator` instance.
+
+By default, the generated class is `public` and _synthetic_ (as defined by the Java Language Specification).
+If you need to make it `final` then use the `ClassCreator.Builder#setFinal()` method.
+
+The `ClassCreator.Builder#superClass()` and ``ClassCreator.Builder#interfaces()` methods can be used to specify the superclass and implemented interfaces.
+
+== Adding Fields
+
+Once we have a `ClassCreator` instance we can start adding fields and methods.
+Fields can be added via the `FieldCreator` interface returned from the `ClassCreator#getFieldCreator()` methods.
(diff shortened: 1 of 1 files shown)


USAGE.adoc
```diff
@@ -0,0 +1,255 @@
...
+<1> Create a method with the following signature `private boolean alwaysReturnFalse()`.
+<2> A `MethodCreator` must always define a return statement.
+<3> This is an equivalent of `return false;`.
+
+[[method_bytecode]]
+== Generating Method Bytecode
+
+The `BytecodeCreator` interface allows you to generate some most common bytecode operations.
+It's not intended to cover all the functionality provided by JVM and Java language though.
+Most operations return and accept a `ResultHandle`.
```
mkouba, line 76: I'm not so sure about this description of `ResultHandle` but I didn't find a better words... so feel free to reword my initial attempt.

USAGE.adoc
```diff
@@ -0,0 +1,255 @@
...
+  myField.setModifiers(ACC_FINAL | ACC_PROTECTED); <2>
+}
+----
+<1> Creates a field `score` of type `Integer`.
+<2> Makes the field `protected final`.
+
+== Adding Methods
+
+A method can be added through a `MethodCreator` interface returned from the `ClassCreator#getMethodCreator()` methods.
+By default, the generated field is `public`.
```
Ladicek, line 49:
```suggestion
By default, the generated method is `public`.
```

USAGE.adoc
```diff
@@ -0,0 +1,255 @@
...
+You can use the `MethodCreator#setModifiers()` method to control the modifiers of the generated field.
```
Ladicek, line 50:
```suggestion
You can use the `MethodCreator#setModifiers()` method to control the modifiers of the generated method.
```

USAGE.adoc
Ladicek: I think your explanation makes sense, even if it is factually incorrect.

(Gizmo allocates a local variable for almost all `ResultHandle`s, with the exception of a top-of-stack value that is a direct input to the immediately following operation, which is kept on stack and not spilled into a local.)

USAGE.adoc
```diff
@@ -0,0 +1,255 @@
...
+void iterate(MethodCreator method, ResultHandle iterable) {
+    // for (Object element : iterable) { 
+    //   System.out.println(element);
+    // }
+    ForEachLoop loop = method.forEach(list);
+    BytecodeCreator block = loop.block();
+    Gizmo.systemOutPrintln(block, loop.element());
+}
+----
+
```
Ladicek, line 185: Maybe add a note here that `ForEachLoop` only works for `Iterable`s and not for arrays?

Ladicek (approved):
Very nice!

Sanne:
Love it, thank you! I only need gizmo rarely, and need to find examples each time as w/o practice I forget.. so this is very useful to me.

PR #126, opened 2022-10-07
Don't avoid using TCCL for classes under the `java.` package prefix

zakkak:
Classes under `java.` are not always guaranteed to be visible to the current bytecode, thus we should still use TCCL to be on the safe side.

According to the documentation of `loadClassFromTCCL`:

https://github.com/quarkusio/gizmo/blob/091fe32bf498650cc0278f9f51e96bdb901ef18b/src/main/java/io/quarkus/gizmo/BytecodeCreator.java#L338-L339

On the other hand the documentation of `loadClass` makes clear that it requires the class being loaded to be accessible from the generated bytecode:

https://github.com/quarkusio/gizmo/blob/091fe32bf498650cc0278f9f51e96bdb901ef18b/src/main/java/io/quarkus/gizmo/BytecodeCreator.java#L299-L301

This PR:

* Adds tests loading a private class using both methods.
* Reverts: https://github.com/quarkusio/gizmo/commit/091fe32bf498650cc0278f9f51e96bdb901ef18b
* Fixes: https://github.com/quarkusio/quarkus/issues/28431

### DIFF
--- a/src/main/java/io/quarkus/gizmo/BytecodeCreatorImpl.java
+++ b/src/main/java/io/quarkus/gizmo/BytecodeCreatorImpl.java
@@ -352,7 +352,7 @@ class BytecodeCreatorImpl implements BytecodeCreator {
         Objects.requireNonNull(className);
         final Class<?> primitiveType = matchPossiblyPrimitive(className);
         if (primitiveType == null) {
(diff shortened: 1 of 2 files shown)


Sanne (approved):
Thanks for the quick fix!

PR #131, opened 2022-10-18
Bitwise operations

Ladicek:


src/main/java/io/quarkus/gizmo/Gizmo.java
```diff
@@ -1110,13 +1113,13 @@ private void generateEquals() {
                                 .falseBranch().returnBoolean(false);
                         break;
                     case "F": // float
-                        // could use `Float.compare` instead of `==`
-                        equals.ifZero(equals.compareFloat(thisValue, thatValue, false))
+                        // this is consistent with Arrays.equals() and it's also what IntelliJ generates
+                        equals.ifIntegerEqual(equals.invokeStaticMethod(FLOAT_TO_INT_BITS, thisValue), equals.invokeStaticMethod(FLOAT_TO_INT_BITS, thatValue))
```
dmlloyd, line 1117: It's pretty wild that `Arrays.equals()` is doing bitwise equality. 🤯

Importantly, this also makes this method consistent with `Float.hashCode()` which is doing `floatToIntBits` as well.

dmlloyd (approved):

stuartwdouglas (approved):

src/main/java/io/quarkus/gizmo/Gizmo.java
Ladicek: Ha, I didn't even notice what `Float|Double.hashCode()` does, good point.

gsmet:
@Ladicek do you have anything else planned or should I ship a release?

Ladicek:
@gsmet if you could do a release now, that would be great! I have a few changes for Quarkus waiting for my recent Gizmo improvements.

I do have some more improvements in mind, but those don't block any immediate Quarkus work and are longer term anyway.

gsmet:
1.4.0.Final should be available in Maven Central in ~30 minutes.

PR #134, opened 2022-11-09
Allow generating interfaces 

### LINKED ISSUE #133: Allow generating interfaces
`ClassCreator` currently calls ASM ClassVisitor with ` ACC_PUBLIC | ACC_SUPER | ACC_SYNTHETIC | extraAccess` around here:
https://github.com/quarkusio/gizmo/blob/f77f4526ad2822ef40d9c5a6b7fb732eb17ffcf8/src/main/java/io/quarkus/gizmo/ClassCreator.java#L177

which is only suitable for common classes, but not interfaces. To make interfaces possible, we need

* A way to set `ACC_INTERFACE` and `ACC_ABSTRACT` flags
* A way to unset `ACC_SUPER`

ppalaga:
Fix #133

This does what I need, but given that I am not a bytecode guru I am not sure at all this is not too straightforward.

### DIFF
--- a/src/main/java/io/quarkus/gizmo/ClassCreator.java
+++ b/src/main/java/io/quarkus/gizmo/ClassCreator.java
@@ -49,7 +49,7 @@ public class ClassCreator implements AutoCloseable, AnnotatedElement, SignatureE
     private final BytecodeCreatorImpl enclosing;
     private final ClassOutput classOutput;
     private final String superClass;
-    private final int extraAccess;
+    private final int access;
     private final String[] interfaces;
     private final Map<MethodDescriptor, MethodCreatorImpl> methods = new LinkedHashMap<>();
     private final Map<FieldDescriptor, FieldCreatorImpl> fields = new LinkedHashMap<>();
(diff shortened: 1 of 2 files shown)


mkouba:
I think that it looks good!

CC @Ladicek @Sanne @geoand

src/main/java/io/quarkus/gizmo/ClassCreator.java
```diff
@@ -313,9 +313,20 @@ public Builder superClass(Class<?> superClass) {
...
-                extraAccess |= Opcodes.ACC_FINAL;
+                access |= Opcodes.ACC_FINAL;
             } else {
-                extraAccess &= ~Opcodes.ACC_FINAL;
+                access &= ~Opcodes.ACC_FINAL;
+            }
+            return this;
+        }
+
+        public Builder setInterface(boolean isInterface) {
```
geoand, line 323: Do we really need the boolean parameter? I mean, is anyone ever going to call this with `false`?

mkouba: Good point.

Ladicek: I think it would be better to have `ClassCreator.interfaceBuilder()` (in addition to `ClassCreator.builder()`), and get rid of `setInterface()` altogether.

src/main/java/io/quarkus/gizmo/ClassCreator.java
```diff
@@ -313,9 +313,20 @@ public Builder superClass(Class<?> superClass) {
 
         public Builder setFinal(boolean isFinal) {
```
Ladicek, line 314: This method should probably throw if we're constructing an interface?

src/main/java/io/quarkus/gizmo/ClassCreator.java
mkouba: Sounds good but wouldn't it make sense to introduce `InterfaceCreator` as well? I mean `ClassCreator.Builder.superClass()` methods do not make sense either.

Ladicek: Yeah, there's probably a few methods on `ClassCreator.Builder` that don't make sense for interfaces. I've commented on `setFinal`, `superClass` is probably also one of them.

Ladicek: I've implemented my ideas here: https://github.com/Ladicek/gizmo/commits/interfaces -- it's this PR's commit amended, so I could just push to @ppalaga's branch if that's OK.

ppalaga: Thanks @Ladicek, your proposal looks great! Let me close this one and let's go with a new PR by you.

ppalaga:
Closing in favor of @Ladicek 's https://github.com/Ladicek/gizmo/commits/interfaces

Ladicek:
#140

PR #136, opened 2022-11-15
Do not attempt to load JDK classes from the TCCL

geoand:
This is done because there is no point involving the TCCL to load JDK classes, since these classes can never be loaded by custom classloaders.

I came across this while looking into the generated code of the Reactive REST Client where we can see a lot of:

```java
         ClassLoader var3 = Thread.currentThread().getContextClassLoader();
         Class var4 = Class.forName("java.lang.String", false, var3);
```

which is completely pointless

### DIFF
--- a/src/main/java/io/quarkus/gizmo/BytecodeCreatorImpl.java
+++ b/src/main/java/io/quarkus/gizmo/BytecodeCreatorImpl.java
@@ -352,14 +352,19 @@ class BytecodeCreatorImpl implements BytecodeCreator {
         final Class<?> primitiveType = matchPossiblyPrimitive(className);
         if (primitiveType == null) {
             if (useTccl) {
-                if (cachedTccl == null) {
-                    ResultHandle currentThread = invokeStaticMethod(THREAD_CURRENT_THREAD);
-                    cachedTccl = invokeVirtualMethod(THREAD_GET_TCCL, currentThread);
+                if (!isNonPrimitiveSafeJavaClass(className)) {
+                    if (cachedTccl == null) {
+                        ResultHandle currentThread = invokeStaticMethod(THREAD_CURRENT_THREAD);
+                        cachedTccl = invokeVirtualMethod(THREAD_GET_TCCL, currentThread);
+                    }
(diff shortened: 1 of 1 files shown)


src/main/java/io/quarkus/gizmo/BytecodeCreatorImpl.java
```diff
@@ -352,14 +352,19 @@ private ResultHandle loadClass(final String className, final boolean useTccl) {
         final Class<?> primitiveType = matchPossiblyPrimitive(className);
         if (primitiveType == null) {
             if (useTccl) {
-                if (cachedTccl == null) {
-                    ResultHandle currentThread = invokeStaticMethod(THREAD_CURRENT_THREAD);
-                    cachedTccl = invokeVirtualMethod(THREAD_GET_TCCL, currentThread);
+                if (isNonPrimitiveJavaClass(className)) {
```
gsmet, line 355: Should it be `! `? Or I misunderstood your intent.

Also I think it was done at some point and @zakkak reverted it: https://github.com/quarkusio/gizmo/pull/126 .

geoand: You are right, the check was wrong.

geoand: And indeed https://github.com/quarkusio/gizmo/pull/126 seems to cause a problem, so let me close this

geoand: Actually, I'll reopen this, but limit ths scope

geoand: PR updated

gsmet (approved):
Looks fine from my side now but let's wait for @stuartwdouglas and @zakkak before merging.

cescoffier (approved):

gsmet:
@cescoffier hmmm. I said we should wait for @zakkak 's and @stuartwdouglas 's approval :).

PR #137, opened 2022-11-15
Revert "Do not attempt to load JDK classes from the TCCL"

zakkak:
@geoand unfortunately #126 results in https://github.com/quarkusio/quarkus/issues/28431

I have added a new test to for `java.lang` that demonstrates this.


geoand (approved):
Thanks for the test

gsmet:
Wouldn't it be possible to not use the TCCL but keep the `Class.forName` approach?

Not sure if it would buy us anything though.

geoand:
What I'll probably end up doing is just keeping a list of classes were we can safely do what I did in the previous PR

geoand:
https://github.com/quarkusio/gizmo/pull/138 is what I have in mind. Not pretty, but should cover most of the usages we have in Quarkus

PR #138, opened 2022-11-16
Do not attempt to load certain JDK classes from the TCCL

geoand:


geoand:
So not pretty.... But it least makes a lot of the Quarkus generated code more sane :)

PR #139, opened 2022-11-22
Require non-null result handles for array agrs

mkouba:
- also require non-null result handles for some array operations


Ladicek (approved):
I'd probably make the `requireNonNullHandles` method `static` (and maybe move to another class), but LGTM overall.

mkouba:
> I'd probably make the `requireNonNullHandles` method `static` (and maybe move to another class), but LGTM overall.

Well, this only makes sense if there are other classes that could reuse such a util method. And I don't see any such class atm. That said, once we have such a class I'd be happy to extract the method.

PR #140, opened 2022-11-22
Support generating interfaces

### LINKED ISSUE #133: Allow generating interfaces

Ladicek:
Fixes #133

### DIFF
--- a/src/main/java/io/quarkus/gizmo/ClassCreator.java
+++ b/src/main/java/io/quarkus/gizmo/ClassCreator.java
@@ -16,6 +16,8 @@
 
 package io.quarkus.gizmo;
 
+import static org.objectweb.asm.Opcodes.ACC_ABSTRACT;
+import static org.objectweb.asm.Opcodes.ACC_INTERFACE;
 import static org.objectweb.asm.Opcodes.ACC_PUBLIC;
 import static org.objectweb.asm.Opcodes.ACC_SUPER;
 import static org.objectweb.asm.Opcodes.ACC_SYNTHETIC;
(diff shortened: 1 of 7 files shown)


src/main/java/io/quarkus/gizmo/ClassCreator.java
```diff
@@ -77,6 +86,10 @@ public ClassCreator(ClassOutput classOutput, String name, String signature, Stri
     }
 
     public MethodCreator getMethodCreator(MethodDescriptor methodDescriptor) {
+        if (this.isInterface() && "<init>".equals(methodDescriptor.getName())) {
```
mkouba, line 89: `"<init>"` would definitely deserve a constant... that could be reused by other gizmo classes as well. It's a pity that `org.jboss.jandex.Utils.INIT_METHOD_NAME` is not public.

ppalaga (approved):

mkouba (approved):

gsmet:
Let me know if a release is needed.

PR #143, opened 2022-12-21
Add the StringSwitch and EnumSwitch constructs

mkouba:
This pull request introduces the `StringSwitch` and `EnumSwitch` constructs that can be used to generate bytecode similar to the Java switch.

API usage examples:

```java
// switch(someString) {
//     case "boom":
//     case "foo":
//          return "fooo";
//     case "bar":
//          return "barr"
//      default:
//          return null;
// }
StringSwitch s = method.stringSwitch(someStringResultHandle);
s.caseOf(List.of("boom", "foo"), bc -> {
  bc.returnValue(bc.load("fooo"));
});
s.caseOf("bar", bc -> {
  bc.returnValue(bc.load("barr"));
});
s.defaultCase(bc -> bc.returnNull());
```

```java
// switch(status) {
//     case ON:
//     case OFF:
//          return status.toString();
//     case UNKNOWN:
//          return "?";
//      default:
//          return null;
// }
EnumSwitch<Status> s = method.enumSwitch(method.getMethodParam(0), Status.class);
s.caseOf(List.of(Status.ON, Status.OFF), bc -> {
   bc.returnValue(Gizmo.toString(bc, method.getMethodParam(0)));
});
s.caseOf(Status.UNKNOWN, bc -> {
   bc.returnValue(bc.load("?"));
});
s.defaultCase(bc -> bc.returnNull());
```

The `StringSwitch` is implemented with the `lookupswitch` instruction while the `EnumSwitch` is implemented with the `tableswitch` instruction. In general, the resulting bytecode is very similar to the bytecode generated by `javac`.

We could also introduce a switch for `int` but enums and strings are probably more common.

### DIFF
--- a/src/main/java/io/quarkus/gizmo/BytecodeCreator.java
+++ b/src/main/java/io/quarkus/gizmo/BytecodeCreator.java
@@ -1034,6 +1034,24 @@ public interface BytecodeCreator extends AutoCloseable {
     }
 
     /**
+     * Create a new switch construct for a string value.
+     * 
+     * @param value The string value to switch on
+     * @return the switch construct
+     */
+    Switch.StringSwitch stringSwitch(ResultHandle value);
+
+    /**
+     * Create a new switch construct for an enum constant.
+     * 
+     * @param <E>
+     * @param value The enum constant to switch on
+     * @param enumClass
+     * @return the switch construct
+     */
+    <E extends Enum<E>> Switch.EnumSwitch<E> enumSwitch(ResultHandle value, Class<E> enumClass);
+
+    /**
      * Indicate that the scope is no longer in use. The scope may refuse additional instructions after this method
      * is called.
      */
--- a/src/main/java/io/quarkus/gizmo/StringSwitchImpl.java
+++ b/src/main/java/io/quarkus/gizmo/StringSwitchImpl.java
@@ -1,0 +1,144 @@
+package io.quarkus.gizmo;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Iterator;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Map.Entry;
+import java.util.Objects;
+import java.util.Set;
+import java.util.function.Consumer;
+
+import org.objectweb.asm.Label;
+import org.objectweb.asm.MethodVisitor;
+
+class StringSwitchImpl extends AbstractSwitch<String> implements Switch.StringSwitch {
+    
+    private final Map<Integer, List<Entry<String, Consumer<BytecodeCreator>>>> hashToCaseBlocks;
+
+    public StringSwitchImpl(ResultHandle value, BytecodeCreatorImpl enclosing) {
+        super(enclosing);
+        this.hashToCaseBlocks = new LinkedHashMap<>();
+        ResultHandle strHash = invokeVirtualMethod(MethodDescriptor.ofMethod(Object.class, "hashCode", int.class), value);
+
+        Set<ResultHandle> inputHandles = new HashSet<>();
+        inputHandles.add(value);
+        inputHandles.add(strHash);
+
+        operations.add(new Operation() {
+
+            @Override
+            void writeBytecode(MethodVisitor methodVisitor) {
+                Map<Integer, Label> hashToLabel = new HashMap<>();
+                List<BytecodeCreatorImpl> lookupBlocks = new ArrayList<>();
+                Map<String, BytecodeCreatorImpl> caseBlocks = new LinkedHashMap<>();
+                BytecodeCreatorImpl defaultBlock = new BytecodeCreatorImpl(StringSwitchImpl.this);
+                if (defaultBlockConsumer != null) {
+                    defaultBlockConsumer.accept(defaultBlock);
+                }
+
+                // Initialize the case blocks and lookup blocks
+                for (Entry<Integer, List<Entry<String, Consumer<BytecodeCreator>>>> hashEntry : hashToCaseBlocks.entrySet()) {
+                    BytecodeCreatorImpl lookupBlock = new BytecodeCreatorImpl(StringSwitchImpl.this);
+                    for (Entry<String, Consumer<BytecodeCreator>> caseEntry : hashEntry.getValue()) {
+                        BytecodeCreatorImpl caseBlock = new BytecodeCreatorImpl(StringSwitchImpl.this);
+                        Consumer<BytecodeCreator> blockConsumer = caseEntry.getValue();
+                        blockConsumer.accept(caseBlock);
+                        if (blockConsumer != EMPTY_BLOCK && !fallThrough) {
+                            caseBlock.breakScope(StringSwitchImpl.this);
+                        }
+                        caseBlock.findActiveResultHandles(inputHandles);
+                        caseBlocks.put(caseEntry.getKey(), caseBlock);
+                        BytecodeCreatorImpl isEqual = (BytecodeCreatorImpl) lookupBlock
+                                .ifTrue(Gizmo.equals(lookupBlock, lookupBlock.load(caseEntry.getKey()), value)).trueBranch();
+                        isEqual.jumpTo(caseBlock);
+                    }
+                    hashToLabel.put(hashEntry.getKey(), lookupBlock.getTop());
+                    lookupBlock.findActiveResultHandles(inputHandles);
+                    lookupBlocks.add(lookupBlock);
+                }
+
+                // Load the hashCode of the tested value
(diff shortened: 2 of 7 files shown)


src/main/java/io/quarkus/gizmo/BytecodeCreator.java
```diff
@@ -946,6 +947,14 @@ default void continueScope() {
      */
     void breakScope(BytecodeCreator scope);
 
+    /**
+     * Go the the top of the given scope. Unlike {@link #continueScope(BytecodeCreator)} this method does not verify if this
+     * bytecode creator is scoped within the given bytecode creator.
+     * 
+     * @param scope
+     */
+    void jumpTo(BytecodeCreator scope);
```
dmlloyd, line 956: This is tricky because by breaking scoping, we get rid of an easy way to reduce our local variable usage (by limiting the lifetime of a value to its containing scope). There are other ways to do it if we decide to later, but they involve liveness tracking which is not very simple.

src/main/java/io/quarkus/gizmo/BytecodeCreator.java
```diff
@@ -1033,6 +1042,25 @@ default ResultHandle increment(ResultHandle toIncrement) {
         return add(toIncrement, load(1));
     }
 
+    /**
+     * Create a new switch construct.
+     * 
+     * @param value The string value to switch on
+     * @param defaultBlockConsumer
```
dmlloyd, line 1049: Missing doc; what is this argument for?

src/main/java/io/quarkus/gizmo/StringSwitchImpl.java
```diff
@@ -0,0 +1,120 @@
...
+                for (Entry<Integer, List<Entry<String, Consumer<BytecodeCreator>>>> hashEntry : hashToCaseBlocks.entrySet()) {
+                    BytecodeCreatorImpl lookupBlock = new BytecodeCreatorImpl(StringSwitchImpl.this);
+                    for (Entry<String, Consumer<BytecodeCreator>> caseEntry : hashEntry.getValue()) {
+                        BytecodeCreatorImpl caseBlock = new BytecodeCreatorImpl(StringSwitchImpl.this);
+                        caseEntry.getValue().accept(caseBlock);
+                        caseBlock.findActiveResultHandles(inputHandles);
+                        caseBlocks.put(caseEntry.getKey(), caseBlock);
+                        BytecodeCreator isEqual = lookupBlock
+                                .ifTrue(Gizmo.equals(lookupBlock, lookupBlock.load(caseEntry.getKey()), value)).trueBranch();
+                        isEqual.jumpTo(caseBlock);
```
dmlloyd, line 52: Wouldn't it be better to create the case block as the child of the `if`? (I don't remember if that is possible...)

src/main/java/io/quarkus/gizmo/BytecodeCreator.java
mkouba: This argument is used to define the bytecode for the `default` label of the switch. But I'm going to change the API and add something like `StringSwitch#defaultCase()` instead.

src/main/java/io/quarkus/gizmo/BytecodeCreator.java
mkouba: Ok. We don't need to make it part of the public API but add a package-private method on `BytecoreCreatorImpl` instead... and use it only in special cases like this. WDYM?

src/main/java/io/quarkus/gizmo/StringSwitchImpl.java
mkouba: I'm not sure. TBH I've tried to mimic the behavior of `javac` because I would expect that it's a reasonable approach...

Ladicek:
I didn't review properly (yet), but I've got 2 questions about the API:

1. What about collapsing all the `case` values for a single "branch" to one method call? I know that

```java
s.caseOf("boom");
s.caseOf("foo", bc -> {
  bc.returnValue(bc.load("fooo"));
});
```

looks closer to the corresponding Java source code, but wouldn't

```java
s.caseOf(Set.of("boom", "foo"), bc -> {
  bc.returnValue(bc.load("fooo"));
});
```

actually be a better API?

2. How about we ditch fall through? At least by default. Falling through is almost never what you want, and having to explicitly add the `break` statement is a chore. In my opinion, `s.caseOf(...)` should _not_ fall through, and if we want to support it, we should have an extra method that _opts into_ the fall through behavior (e.g. `s.caseWithFallThrough(...)` or so).

mkouba:
> I didn't review properly (yet), but I've got 2 questions about the API:
> 
>     1. What about collapsing all the `case` values for a single "branch" to one method call? I know that
> 
> 
> ```java
> s.caseOf("boom");
> s.caseOf("foo", bc -> {
>   bc.returnValue(bc.load("fooo"));
> });
> ```
> 
> looks closer to the corresponding Java source code, but wouldn't
> 
> ```java
> s.caseOf(Set.of("boom", "foo"), bc -> {
>   bc.returnValue(bc.load("fooo"));
> });
> ```
> 
> actually be a better API?

+1 

But I wonder if we need to keep the ordering or not..? 

> 
>     2. How about we ditch fall through? At least by default. Falling through is almost never what you want, and having to explicitly add the `break` statement is a chore. In my opinion, `s.caseOf(...)` should _not_ fall through, and if we want to support it, we should have an extra method that _opts into_ the fall through behavior (e.g. `s.caseWithFallThrough(...)` or so).

I'm not so sure it's reasonable to "break" by default. I mean it's not "expected".

Ladicek:
I think the `case` order doesn't matter (I mean the order of constants in the Java source code / Gizmo API calls), the generated bytecode is always ordered numerically (at least for `tableswitch`, and `lookupswitch` has some ordering in there too IIRC). But I wrote `Set` without really thinking about it -- it could just as well be `List`.

Fall through is a contentious topic, because Java has it and we should be close to Java, but my Strong Opinion here is that Java should have never copied this from C(++) and Go did the Right Thing (https://golangbot.com/switch/#fallthrough). I guess I don't mind _too much_ though.

mkouba:
> and `lookupswitch` has some ordering in there too IIRC

Hm, you're right that the `lookupswitch` keys must be sorted in increasing numerical order and we use the hashcode as the key. So in theory it should not really matter.

dmlloyd:
> I'm not so sure it's reasonable to "break" by default. I mean it's not "expected".

Java 17's new switch syntax works in this way FWIW.

mkouba:
> > I'm not so sure it's reasonable to "break" by default. I mean it's not "expected".
> 
> Java 17's new switch syntax works in this way FWIW.

Hm, you mean the `case L ->` syntax. Fair enough. 

We could just provide two variants of the method. One for the "switch labeled statement group" (`case L:`) and one for the "switch rule block" (`case L ->`). 

We could even add the `fallThrough()` method to the `Switch` to control the default behavior. Because otherwise we would have to check if different kinds of case blocks are used etc. This way we could keep one method for case blocks.

```java
// equivalent of case L ->
StringSwitch s1 = method.stringSwitch(val);
s1.caseOf("foo", bc -> bc.assign(ret, bc.load("fooo")));
s1.caseOf("bar", bc -> bc.assign(ret, bc.load("barr")));

// equivalent of case L:
StringSwitch s2 = method.stringSwitch(val).fallThrough();
s2.caseOf("foo", bc -> bc.assign(ret, bc.load("fooo")));
s2.caseOf("bar", bc -> bc.assign(ret, bc.load("barr")));
```
@Ladicek @dmlloyd WDYT?

mkouba:
I've added the `Switch.fallThrough()` method.

dmlloyd (approved):

Ladicek:
Sorry it took me a while, I had to browse through this in my IDE and look at the generated bytecode to be able to understand what's going on. LGTM.

Ladicek:
(I also originally wanted to say that perhaps the `fallThrough()` method should be `withFallThrough()` and return a new `Switch` instance, because the current setup allows changing the "fall through" flag in between cases, which is pretty terrible, but I see that wouldn't be super-easy to do, so... I don't mind.)

mkouba:
> Sorry it took me a while, I had to browse through this in my IDE and look at the generated bytecode to be able to understand what's going on. LGTM.

No problem.

> (I also originally wanted to say that perhaps the fallThrough() method should be withFallThrough() and return a new Switch instance, because the current setup allows changing the "fall through" flag in between cases, which is pretty terrible, but I see that wouldn't be super-easy to do, so... I don't mind.)

Well, the `Switch` construct is mutable and it's not intended to be reused. Also you can't mix the syntax (`L->` and `L:`) in Java so I think that it's fine.

mkouba:
FYI I need to send a follow-up PR to fix the problem when using result handles from the enclosing scope.

PR #144, opened 2023-01-04
String/Enum switch -  make result handles from enclosing scope accesible

mkouba:
Follows up on https://github.com/quarkusio/gizmo/pull/143.


Ladicek (approved):

mkouba:
> I was wondering why did the previous version call `findActiveResultHandles`, but didn't quite realize it should override the method instead. Good catch!

To be honest I'm still a bit unsure but the tests pass so it should be fine ;-)

PR #145, opened 2023-01-04
Fix backward compatibility issue with public constructor of ClassCreator

mkouba:
- follows-up on https://github.com/quarkusio/gizmo/pull/140


Ladicek (approved):
Whoops! Thanks!

PR #146, opened 2023-01-04
StringSwitch - fix hashCode collisons

mkouba:
- the previous impl should be functionally equivalent but less efficient if a tested value has the same hashCode as one of the labels but none of the labels is equal to the tested value


mkouba:
@Ladicek I'm sorry but I have one more fix related to the `StringSwitch` ;-)

Ladicek (approved):

Ladicek:
You're keeping me busy :laughing:

PR #147, opened 2023-01-06
Introduce SignatureBuilder API

mkouba:
This pull request supersedes https://github.com/quarkusio/gizmo/pull/116.

Usage example:
```java
import io.quarkus.gizmo.SignatureBuilder;
import io.quarkus.gizmo.Type;

// List<String> test();
method.setSignature(SignatureBuilder.forMethod()
   .setReturnType(Type.parameterizedType(Type.classType(List.class), Type.classType(String.class)))
   .build()
)
```

See also `io.quarkus.gizmo.SignaturesTest`.

### DIFF
--- a/src/main/java/io/quarkus/gizmo/ClassSignatureBuilderImpl.java
+++ b/src/main/java/io/quarkus/gizmo/ClassSignatureBuilderImpl.java
@@ -1,0 +1,93 @@
+package io.quarkus.gizmo;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import org.jboss.jandex.DotName;
+
+import io.quarkus.gizmo.SignatureBuilder.ClassSignatureBuilder;
+import io.quarkus.gizmo.Type.ClassType;
+import io.quarkus.gizmo.Type.ParameterizedType;
+import io.quarkus.gizmo.Type.TypeVariable;
+
+class ClassSignatureBuilderImpl implements ClassSignatureBuilder {
+
+    private Type superClass = Type.classType(DotName.OBJECT_NAME);
+    private List<TypeVariable> typeParameters = new ArrayList<>();
+    private List<Type> superInterfaces = new ArrayList<>();
+
+    @Override
+    public String build() {
+        StringBuilder signature = new StringBuilder();
+
+        // type params
+        if (!typeParameters.isEmpty()) {
+            signature.append('<');
+            for (TypeVariable typeParameter : typeParameters) {
+                typeParameter.appendTypeParameterToSignature(signature);
+            }
+            signature.append('>');
+        }
+
+        // superclass
+        signature.append(superClass.toSignature());
+
+        // interfaces
+        if (!superInterfaces.isEmpty()) {
+            for (Type superInterface : superInterfaces) {
+                signature.append(superInterface.toSignature());
+            }
+        }
+        return signature.toString();
+    }
+
+    @Override
+    public ClassSignatureBuilder addTypeParameter(TypeVariable typeParameter) {
+        typeParameters.add(typeParameter);
+        return this;
+    }
+
+    @Override
+    public ClassSignatureBuilder setSuperClass(ClassType superClass) {
+        this.superClass = superClass;
+        return this;
+    }
+
+    @Override
+    public ClassSignatureBuilder setSuperClass(ParameterizedType superClass) {
+        if (containsWildcard(superClass)) {
+            throw new IllegalArgumentException("A super type may not specify a wilcard");
+        }
+        this.superClass = superClass;
+        return this;
+    }
+
+    @Override
+    public ClassSignatureBuilder addSuperInterface(ClassType interfaceType) {
+        superInterfaces.add(interfaceType);
+        return this;
+    }
+
+    @Override
+    public ClassSignatureBuilder addSuperInterface(ParameterizedType interfaceType) {
+        if (containsWildcard(interfaceType)) {
+            throw new IllegalArgumentException("A super type may not specify a wilcard");
+        }
+        superInterfaces.add(interfaceType);
+        return this;
+    }
+
+    private boolean containsWildcard(Type type) {
+        if (type.isWildcard()) {
+            return true;
+        } else if (type.isParameterizedType()) {
+            for (Type typeArgument : type.asParameterizedType().typeArguments) {
+                if (containsWildcard(typeArgument)) {
+                    return true;
+                }
+            }
+        }
+        return false;
+    }
+
+}
--- a/src/main/java/io/quarkus/gizmo/ClassSignatureBuilderImpl.java
+++ b/src/main/java/io/quarkus/gizmo/ClassSignatureBuilderImpl.java
@@ -9,9 +9,10 @@ import io.quarkus.gizmo.Type.ParameterizedType;
 import io.quarkus.gizmo.Type.TypeVariable;
 
 class ClassSignatureBuilderImpl implements ClassSignatureBuilder {
-    private List<TypeVariable> typeParameters = new ArrayList<>();
-    private Type superClass = ClassType.OBJECT;
-    private List<Type> superInterfaces = new ArrayList<>();
+    
+    List<TypeVariable> typeParameters = new ArrayList<>();
+    Type superClass = ClassType.OBJECT;
+    List<Type> superInterfaces = new ArrayList<>();
 
     @Override
     public String build() {
--- a/src/main/java/io/quarkus/gizmo/ClassSignatureBuilderImpl.java
+++ b/src/main/java/io/quarkus/gizmo/ClassSignatureBuilderImpl.java
@@ -61,13 +61,13 @@ class ClassSignatureBuilderImpl implements ClassSignatureBuilder {
     }
 
     @Override
-    public ClassSignatureBuilder addSuperInterface(ClassType interfaceType) {
+    public ClassSignatureBuilder addInterface(ClassType interfaceType) {
         superInterfaces.add(interfaceType);
         return this;
     }
 
     @Override
-    public ClassSignatureBuilder addSuperInterface(ParameterizedType interfaceType) {
+    public ClassSignatureBuilder addInterface(ParameterizedType interfaceType) {
         if (containsWildcard(interfaceType)) {
             throw new IllegalArgumentException("A super type may not specify a wilcard");
         }
--- a/src/main/java/io/quarkus/gizmo/ClassSignatureBuilderImpl.java
+++ b/src/main/java/io/quarkus/gizmo/ClassSignatureBuilderImpl.java
@@ -3,17 +3,14 @@ package io.quarkus.gizmo;
 import java.util.ArrayList;
 import java.util.List;
 
-import org.jboss.jandex.DotName;
-
 import io.quarkus.gizmo.SignatureBuilder.ClassSignatureBuilder;
 import io.quarkus.gizmo.Type.ClassType;
 import io.quarkus.gizmo.Type.ParameterizedType;
 import io.quarkus.gizmo.Type.TypeVariable;
 
 class ClassSignatureBuilderImpl implements ClassSignatureBuilder {
-
-    private Type superClass = Type.classType(DotName.OBJECT_NAME);
     private List<TypeVariable> typeParameters = new ArrayList<>();
+    private Type superClass = ClassType.OBJECT;
     private List<Type> superInterfaces = new ArrayList<>();
 
     @Override
@@ -30,12 +27,12 @@ class ClassSignatureBuilderImpl implements ClassSignatureBuilder {
         }
 
         // superclass
-        signature.append(superClass.toSignature());
+        superClass.appendToSignature(signature);
 
         // interfaces
         if (!superInterfaces.isEmpty()) {
             for (Type superInterface : superInterfaces) {
-                signature.append(superInterface.toSignature());
+                superInterface.appendToSignature(signature);
             }
         }
         return signature.toString();
@@ -58,6 +55,7 @@ class ClassSignatureBuilderImpl implements ClassSignatureBuilder {
         if (containsWildcard(superClass)) {
             throw new IllegalArgumentException("A super type may not specify a wilcard");
         }
+
         this.superClass = superClass;
         return this;
     }
@@ -81,7 +79,7 @@ class ClassSignatureBuilderImpl implements ClassSignatureBuilder {
         if (type.isWildcard()) {
             return true;
         } else if (type.isParameterizedType()) {
-            for (Type typeArgument : type.asParameterizedType().typeArguments) {
+            for (Type typeArgument : type.asParameterizedType().getTypeArguments()) {
                 if (containsWildcard(typeArgument)) {
                     return true;
                 }
@@ -89,5 +87,4 @@ class ClassSignatureBuilderImpl implements ClassSignatureBuilder {
         }
         return false;
     }
-
 }
--- a/src/main/java/io/quarkus/gizmo/ClassSignatureBuilderImpl.java
+++ b/src/main/java/io/quarkus/gizmo/ClassSignatureBuilderImpl.java
@@ -9,7 +9,7 @@ import io.quarkus.gizmo.Type.ParameterizedType;
 import io.quarkus.gizmo.Type.TypeVariable;
 
 class ClassSignatureBuilderImpl implements ClassSignatureBuilder {
-    
+
     List<TypeVariable> typeParameters = new ArrayList<>();
     Type superClass = ClassType.OBJECT;
     List<Type> superInterfaces = new ArrayList<>();
@@ -54,7 +54,7 @@ class ClassSignatureBuilderImpl implements ClassSignatureBuilder {
     @Override
     public ClassSignatureBuilder setSuperClass(ParameterizedType superClass) {
         if (containsWildcard(superClass)) {
-            throw new IllegalArgumentException("A super type may not specify a wilcard");
+            throw new IllegalArgumentException("An extended class type may not specify a wildcard");
         }
 
         this.superClass = superClass;
@@ -70,7 +70,7 @@ class ClassSignatureBuilderImpl implements ClassSignatureBuilder {
     @Override
     public ClassSignatureBuilder addInterface(ParameterizedType interfaceType) {
         if (containsWildcard(interfaceType)) {
-            throw new IllegalArgumentException("A super type may not specify a wilcard");
+            throw new IllegalArgumentException("An implemented interface type may not specify a wildcard");
         }
         superInterfaces.add(interfaceType);
         return this;
--- a/src/main/java/io/quarkus/gizmo/SignatureBuilder.java
+++ b/src/main/java/io/quarkus/gizmo/SignatureBuilder.java
@@ -37,9 +37,9 @@ public interface SignatureBuilder {
         
         ClassSignatureBuilder setSuperClass(ParameterizedType superClass);
 
-        ClassSignatureBuilder addSuperInterface(ClassType interfaceType);
+        ClassSignatureBuilder addInterface(ClassType interfaceType);
         
-        ClassSignatureBuilder addSuperInterface(ParameterizedType interfaceType);
+        ClassSignatureBuilder addInterface(ParameterizedType interfaceType);
     }
 
     /**
--- a/src/main/java/io/quarkus/gizmo/SignatureBuilder.java
+++ b/src/main/java/io/quarkus/gizmo/SignatureBuilder.java
@@ -1,0 +1,62 @@
+package io.quarkus.gizmo;
+
+import io.quarkus.gizmo.Type.ClassType;
+import io.quarkus.gizmo.Type.ParameterizedType;
+import io.quarkus.gizmo.Type.TypeVariable;
+
+/**
+ * Builds a signature as defined in JVMS 17, chapter "4.7.9.1. Signatures".
+ * 
+ * @see SignatureElement#setSignature(String)
+ */
+public interface SignatureBuilder {
+
+    static ClassSignatureBuilder forClass() {
+        return new ClassSignatureBuilderImpl();
+    }
+
+    static MethodSignatureBuilder forMethod() {
+        return new MethodSignatureBuilderImpl();
+    }
+
+    static FieldSignatureBuilder forField() {
+        return new FieldSignatureBuilderImpl();
+    }
+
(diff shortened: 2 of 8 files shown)


src/main/java/io/quarkus/gizmo/Type.java
```diff
@@ -0,0 +1,418 @@
...
+    public static PrimitiveType charType() {
+        return PrimitiveType.CHAR;
+    }
+
+    public static ClassType classType(DotName name) {
+        return new ClassType(Objects.requireNonNull(name).toString().replace('.', '/'));
+    }
+
+    public static ClassType classType(String name) {
+        return new ClassType(name);
```
Ladicek, line 62: Maybe `requireNonNull` and also `replace('.', '/')`?

src/main/java/io/quarkus/gizmo/Type.java
```diff
@@ -0,0 +1,418 @@
...
+            return classType(classType.getDeclaringClass().getName().replace('.', '/') + "." + classType.getSimpleName());
+        }
+        return classType(classType.getName().replace('.', '/'));
+    }
+
+    public static ParameterizedType parameterizedType(ClassType classType, Type... typeArguments) {
+        return new ParameterizedType(classType, Arrays.asList(typeArguments));
+    }
+
+    public static ArrayType arrayType(Type component) {
```
Ladicek, line 77: This is not a _component_ type, but an _element_ type. There's no difference between these terms in single-dimensional arrays, but in e.g. `String[][]`, the component type is `String[]`, while the element type is `String`.

src/main/java/io/quarkus/gizmo/Type.java
```diff
@@ -0,0 +1,418 @@
...
+
+    public static WildcardType wildcardTypeUnbounded() {
+        return new WildcardType(ClassType.OBJECT, null);
+    }
+
+    /**
+     * 
+     * @return the signature as defined in JVMS 17, chapter "4.7.9.1. Signatures"
+     */
+    String toSignature();
```
Ladicek, line 109: Instead of, or in addition to, a `String toSignature()` method, I'd probably use `void appendSignature(StringBuilder builder)`. Obtaining a signature of a type in isolation is generally useless, I'd say.

src/main/java/io/quarkus/gizmo/Type.java
mkouba: We have `Objects.requireNonNull()` in the `ClassType` constructor and we don't want to do `replace('.', '/')` because of nested classes; e.g. if you want to use something like `org/acme/Foo<TT;>.Bar`. That's the reason why we take the value as is...

src/main/java/io/quarkus/gizmo/Type.java
```diff
@@ -0,0 +1,418 @@
...
+        @Override
+        public TypeVariable asTypeVariable() {
+            return this;
+        }
+
+    }
+
+    public static class ArrayType implements Type {
+
+        final Type component;
```
Ladicek, line 250: Same as above -- this is not a component type, it's an element type.

src/main/java/io/quarkus/gizmo/Type.java
mkouba: Ok, I got confused by the `jakarta.enterprise.lang.model.types.ArrayType.componentType()` javadoc ;-). By the way, `org.jboss.jandex.ArrayType.component` is then wrong too.

src/main/java/io/quarkus/gizmo/Type.java
```diff
@@ -0,0 +1,418 @@
...
+        return new ClassType(Objects.requireNonNull(name).toString().replace('.', '/'));
+    }
+
+    public static ClassType classType(String name) {
+        return new ClassType(name);
+    }
+
+    public static ClassType classType(Class<?> classType) {
+        Objects.requireNonNull(classType);
+        if (classType.getDeclaringClass() != null) {
```
Ladicek, line 67: What does this do? Is it some special case for nested classes? If so, it only works for one level of nesting, no?

src/main/java/io/quarkus/gizmo/Type.java
mkouba: Ok, we can have both.

Ladicek:
Looks great! I added some minor comments, and I have 2 more general here:

1. I'd generally try to keep most of this private. For example, I don't think we need public `Type.toSignature()` or `Type.isClass()` / `asClass()` etc.
2. I didn't think of this before, but it occured to me when reviewing this. Maybe we could split `TypeVariable` to `TypeParameter` and `TypeVariableArgument` for this particular case? A `TypeParameter` may have bounds, while `TypeArgumentVariable` has just the identifier. (`TypeArgumentVariable` is a terrible name...) This would allow a bit more type safety, e.g. a `ParameterizedType` wouldn't accept a `TypeParameter` as a type argument. Not sure it's worth it.

src/main/java/io/quarkus/gizmo/Type.java
Ladicek: The CDI Lang Model javadoc has the exact same example :-)

The Jandex `ArrayType.component()` is generally an element type, but may sometimes be a component type, which is confusing and probably needs even more documentation than I already added. Jandex has more quirks like this, unfortunately.

src/main/java/io/quarkus/gizmo/Type.java
Ladicek: I didn't think of nested classes before, but I'd guess an explicit API to express a class/parameterized type that is nested in another class/parameterized type is probably better than having to do this.

mkouba:
> 1. I'd generally try to keep most of this private. For example, I don't think we need public `Type.toSignature()` or `Type.isClass()` / `asClass()` etc.

Well, we would have to use some abstract superclass because interfaces cannot declare private members and we use these methods in impls. I don't think that it's a problem.

> 2. I didn't think of this before, but it occured to me when reviewing this. Maybe we could split TypeVariable to TypeParameter and TypeVariableArgument for this particular case? A TypeParameter may have bounds, while TypeArgumentVariable has just the identifier. (TypeArgumentVariable is a terrible name...) This would allow a bit more type safety, e.g. a ParameterizedType wouldn't accept a TypeParameter as a type argument. Not sure it's worth it.

Hmm. I'm not sure. From user's POV I think that the name "type variable" is more understandable. In any case, I believe that the current version is "good enough".

src/main/java/io/quarkus/gizmo/Type.java
mkouba: Yes, nested classes are a bit problematic. And yes, you're right that this is wrong ;-)

mkouba: The use case we don't support at the moment is a nested class declared on parameterized type.

mkouba: I've removed this logic.

Ladicek:
Added one commit that refactors the `SignatureBuilder` for better encapsulation (per my ideas above). Feel free to drop :-)

The tests could use some improvements -- I'll get to it next week.

gsmet:
I will have a look on Monday and try to use it in Quarkus GitHub App.

Sgitario:
Back from my PTO. I will use these changes in some Quarkus extensions to give some feedback asap. Thanks!

src/main/java/io/quarkus/gizmo/Signature.java
```diff
@@ -0,0 +1,49 @@
+package io.quarkus.gizmo;
+
+public interface Signature {
```
mkouba, line 3: I don't see a reason for this interface. Also for every signature you'll create a completely useless object (e.g. `ClassSignature`) and it's also more difficult to use: `Signature.forClass().build().toString()` VS `SignatureBuilder.forClass().build()`...

src/main/java/io/quarkus/gizmo/Signature.java
```diff
@@ -0,0 +1,49 @@
...
+    static FieldSignatureBuilder forField() {
+        return new FieldSignatureBuilder();
+    }
+
+    class ClassSignature implements Signature {
+    }
+
+    class ClassSignatureBuilder {
+        ClassSignature build() {
+            return null;
```
mkouba, line 28: ??

mkouba:
> Added one commit that refactors the `SignatureBuilder` for better encapsulation (per my ideas above). Feel free to drop :-)
> 

@Ladicek I can't say I like all those `(TypeImpl)` casts and similar but other than the `Signature` interface it looks very good.

> The tests could use some improvements -- I'll get to it next week.

Feel free to improve the coverage ;-).

src/main/java/io/quarkus/gizmo/Signature.java
Ladicek: Ugh. I don't recall adding this, I probably had this locally from my previous unfinished experiments with this. It doesn't make any sense, it's the same as `SignatureBuilder`. Will delete.

Ladicek: Done.

Ladicek:
Done.

mkouba:
> Done.

Ok, I think that we're ready to go. We could either merge this PR and improve the coverage in a follow-up or wait a little... WDYT?

gsmet:
Maybe wait for this afternoon? I want to test it on Quarkus GitHub App and report back.

gsmet:
(Working on Sergey's PR this morning so can't work on that now)

mkouba:
> Maybe wait for this afternoon? I want to test it on Quarkus GitHub App and report back.

:+1:

Ladicek:
I did add the most interesting test case I wanted (the horror of inner classes). There's more that could be done for methods, but I feel the code by now is correct by construction. I have a very similar piece of code in Jandex with a lot more tests, so I feel pretty good about what we have.

src/main/java/io/quarkus/gizmo/SignatureBuilder.java
```diff
@@ -0,0 +1,66 @@
...
+import io.quarkus.gizmo.Type.ClassType;
+import io.quarkus.gizmo.Type.ParameterizedType;
+import io.quarkus.gizmo.Type.TypeVariable;
+
+/**
+ * Builds a generic signature as defined in JVMS 17, chapter "4.7.9.1. Signatures".
+ * 
+ * @see SignatureElement#setSignature(String)
+ */
+public interface SignatureBuilder {
```
gsmet, line 12: It's minor but shouldn't this class be a class? It seems weird to me that we make it an interface given we don't really want implementations.

src/main/java/io/quarkus/gizmo/ClassSignatureBuilderImpl.java
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    @Override
+    public ClassSignatureBuilder addTypeParameter(TypeVariable typeParameter) {
+        typeParameters.add(typeParameter);
+        return this;
+    }
+
+    @Override
+    public ClassSignatureBuilder setSuperClass(ClassType superClass) {
```
gsmet, line 48: Given `Class.getSuperclass`, I would rather keep the same case for `Superclass`. Or is it written this way in other parts of Gizmo?

src/main/java/io/quarkus/gizmo/SignatureBuilder.java
mkouba: We have implementations but we don't want users to provide their own implementations which is IMO very common... :shrug:.

src/main/java/io/quarkus/gizmo/ClassSignatureBuilderImpl.java
```diff
@@ -0,0 +1,90 @@
...
+        if (containsWildcard(superClass)) {
+            throw new IllegalArgumentException("A super type may not specify a wilcard");
+        }
+
+        this.superClass = superClass;
+        return this;
+    }
+
+    @Override
+    public ClassSignatureBuilder addSuperInterface(ClassType interfaceType) {
```
gsmet, line 64: Does `SuperInterfaces` make sense? Shouldn't it be `addInterface` or `implementInterface`?

src/main/java/io/quarkus/gizmo/ClassSignatureBuilderImpl.java
```diff
@@ -0,0 +1,90 @@
...
+        superInterfaces.add(interfaceType);
+        return this;
+    }
+
+    @Override
+    public ClassSignatureBuilder addSuperInterface(ParameterizedType interfaceType) {
+        if (containsWildcard(interfaceType)) {
+            throw new IllegalArgumentException("A super type may not specify a wilcard");
```
gsmet, line 72: Maybe this could use some tuning if we adjust `SuperInterface`

src/main/java/io/quarkus/gizmo/ClassSignatureBuilderImpl.java
gsmet: Ah I see that `SuperClass` is used in other parts of Gizmo. That's a bit unfortunate but I suppose it's too late to change it.

gsmet:
Ah I thought this comment was part of my ongoing review and I removed it
when noticing it was actually used as an interface. You can ignore me on
this one :).

src/main/java/io/quarkus/gizmo/ClassSignatureBuilderImpl.java
gsmet: FWIW `ClassCreator` uses `interfaces()` and I wonder if we should stick to the same method names so `superClass()` and `interfaces()`.

gsmet (changes requested):
I added a few comments. I'm far less familiar with bytecode than you are so hopefully they won't be too stupid.

What I'm wondering is if it's the right solution because my code looks like:

```java
                String signature = SignatureBuilder.forClass()
                        .setSuperClass(parameterizedType(classType(AnnotationLiteral.class), classType(eventAnnotationLiteral.getName())))
                        .addSuperInterface(classType(eventAnnotationLiteral.getName()))
                        .build();

                ClassCreator literalClassCreator = ClassCreator.builder().classOutput(classOutput)
                        .className(literalClassName)
                        .signature(signature)
                        .superClass(AnnotationLiteral.class)
                        .interfaces(eventAnnotationLiteral.getName().toString())
                        .build();
```

The code looks a bit redundant with the signature being defined together with the superClass and interfaces.
Can I get rid of them when I define the signature? If not, I'm basically pushing the same information twice and it's a bit error prone.

Other than this comment and the ones inline, it seems to work well in my case.

src/main/java/io/quarkus/gizmo/ClassSignatureBuilderImpl.java
mkouba: I don't understand. Where is it used? And what exactly would you like to keep?

src/main/java/io/quarkus/gizmo/ClassSignatureBuilderImpl.java
mkouba: Well, it's a [jvms language](https://docs.oracle.com/javase/specs/jvms/se11/html/jvms-4.html#jvms-ClassTypeSignature) and it's consistent with `setSuperClass()`. That said, I would not mind if we change the method name to `addInterface()`.

src/main/java/io/quarkus/gizmo/ClassSignatureBuilderImpl.java
mkouba: You mean the message? Hm, so the `SuperInterface` is also a "super type" and the stack trace should contain enough information IMO. But again, if we decide to `A super interface may not specify a wilcard` then OK ;-)

src/main/java/io/quarkus/gizmo/ClassSignatureBuilderImpl.java
gsmet: My point was that in the JDK APIs, it's `Superclass`, not `SuperClass`. But Gizmo is using `SuperClass` in other places so we should probably keep it as is.
Is it clearer?

src/main/java/io/quarkus/gizmo/ClassSignatureBuilderImpl.java
gsmet: I would rather be consistent with what we do with `ClassCreator` as much as we can.

gsmet:
@mkouba have you seen my general comment about the redundancy?

src/main/java/io/quarkus/gizmo/ClassSignatureBuilderImpl.java
mkouba: :+1: I see..

mkouba:
> @mkouba have you seen my general comment about the redundancy?

Yes.

> Can I get rid of them when I define the signature? If not, I'm basically pushing the same information twice and it's a bit error prone.

You can't and yes it's a bit error prone. But we wanted to keep the current `SignatureElement` API where the signature is just a `String`.

@Ladicek WDYT?

src/main/java/io/quarkus/gizmo/ClassSignatureBuilderImpl.java
Ladicek: We could change this error message (and the other one in the `setSuperClass`) a bit I guess, even if it's quite precise as is. Something like "An implemented interface type may not specify a wildcard" or so (similarly, "an extended class type may not specify a wildcard").

src/main/java/io/quarkus/gizmo/ClassSignatureBuilderImpl.java
mkouba: Ok, let's use `addInterface()` then...

Ladicek:
Yeah, I agree this is unfortunate. The signature contains more information in case the supertypes involve generics, so the only way to avoid the duplicity that I can see is to allow passing `SignatureBuilder` (or an immutable `Signature` perhaps) to the `ClassCreator.Builder`  and that would automatically set the signature, the superclass, and the interfaces.

src/main/java/io/quarkus/gizmo/ClassSignatureBuilderImpl.java
gsmet: I know I'm getting annoying but... for the `ClassCreator` builder, we have `superClass()` and `interfaces()` and for this builder, we have `setSuperClass()` and `addInterface()`.
Given the singular `interface()` won't compile, it might be for the best but I wanted to make sure you were are of the inconsistency.

mkouba: Well, `ClassCreator` is not a "real" builder. And `set/add` IMO make the most sense for builder methods (the other option is `with` but I think that `set/add` is more clear).

mkouba: Ah, you're talking about `io.quarkus.gizmo.ClassCreator.Builder`... Hmm.

mkouba:
Passing a `ClassSignatureBuilder` to the `io.quarkus.gizmo.ClassCreator.Builder` seems to be the "best" option but I don't like it very much.

Ladicek:
I think we could just add an overload of `signature()` that would accept `ClassSignatureBuilder` and when that is used, `superClass()` and `interfaces()` become optional. Maybe we could even verify that when both are set, they match.

gsmet:
Another option would be to create a `ClassSignature` from the builder and have `signature(ClassSignature)` as an overload of `signature(String)`.

That would probably be a bit better than passing a builder.

mkouba:
> That would probably be a bit better than passing a builder.

Except that it would require one more useless allocation...

Ladicek:
Also one more [otherwise] useless class. I didn't get to trying it yet (generic signatures and descriptors are so much fun... :-) ), but I suspect using a builder will be just fine.

src/main/java/io/quarkus/gizmo/Type.java
```diff
@@ -0,0 +1,488 @@
...
+
+        /**
+         * Allows building a signature like {@code Lcom/example/Outer.Inner;}. This is usually
+         * unnecessary, because {@code Lcom/example/Outer$Inner} is also a valid signature,
+         * but it's occasionally useful to build a signature of more complex inner types.
+         *
+         * @param simpleName simple name of the member class nested in this class
+         * @return the inner class
+         */
+        public ClassType innerClass(String simpleName) {
```
mkouba, line 279: @Ladicek BTW `innerClass`  is IMO inaccurate because only _non-static nested classes_ are called _inner classes_ in Java... so I think it should be `nestedClass()`.

Ladicek: For some reason, I got pretty convinced this is only useful for inner classes and not for static nested classes. Let me look again.

gsmet (approved):

gsmet:
Didn't want my review to be blocking, I will let you all have fun with it and merge :).

mkouba:
> Didn't want my review to be blocking, I will let you all have fun with it and merge :).

No problem. It's a great feedback. Thanks!

mkouba:
> Also one more [otherwise] useless class. I didn't get to trying it yet (generic signatures and descriptors are so much fun... :-) ), but I suspect using a builder will be just fine.

I have something... will add a commit - feel free to refactor/throw it away.

Ladicek:
No throwing away, this is pretty much what I thought I'd do. I think it's fine.

src/main/java/io/quarkus/gizmo/Type.java
mkouba: Hm, let's keep the `inner` prefix for now...

mkouba:
> Didn't want my review to be blocking, I will let you all have fun with it and merge :).

@gsmet How about something like this? https://github.com/quarkusio/gizmo/pull/147/files#diff-f3ceec916aa72c0ab7318fa5821d400412118071f45e55d13310c5f4a3796df9R260-R274

src/main/java/io/quarkus/gizmo/ClassSignatureBuilderImpl.java
Ladicek: This wasn't addressed IIUC? Not a big deal in my opinion, just noting.

mkouba: Just an oversight... will fix it shortly.

Ladicek (approved):

Ladicek:
Yay! :laughing:

PR #151, opened 2023-04-12
Add support for loading String's larger than 65535

geoand:
Fixes: https://github.com/quarkusio/quarkus/issues/32507

### DIFF
--- a/src/main/java/io/quarkus/gizmo/BytecodeCreatorImpl.java
+++ b/src/main/java/io/quarkus/gizmo/BytecodeCreatorImpl.java
@@ -67,6 +67,7 @@ class BytecodeCreatorImpl implements BytecodeCreator {
                     ArrayList.class.getName(),
                     Map.class.getName(),
                     HashMap.class.getName());
+    static final int MAX_STRING_LENGTH = 65535;
 
     protected final List<Operation> operations = new ArrayList<>();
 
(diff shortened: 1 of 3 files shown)


src/main/java/io/quarkus/gizmo/BytecodeCreatorImpl.java
```diff
@@ -311,11 +312,25 @@ ResultHandle getOutgoingResultHandle() {
...
         Objects.requireNonNull(val);
-        if (val.length() > 65535) {
-            //TODO: we could auto split this, but I don't think we really want strings this long
-            throw new IllegalArgumentException("Cannot load strings larger than " + 65535 + " bytes");
+        int length = val.length();
+        if (length > MAX_STRING_LENGTH) {
+            int div = length / MAX_STRING_LENGTH;
+            int mod = length % MAX_STRING_LENGTH;
+
+            Gizmo.StringBuilderGenerator sb = Gizmo.newStringBuilder(this);
```
stuartwdouglas, line 320: Minor nit, but shouldn't this allocate one with the correct length, as we know how long the resulting string will be?

geoand: Good point, I'll update it

geoand: Done

stuartwdouglas (approved):

src/main/java/io/quarkus/gizmo/BytecodeCreatorImpl.java
```diff
@@ -311,11 +312,25 @@ ResultHandle getOutgoingResultHandle() {
...
+                int endIndex = beginIndex + MAX_STRING_LENGTH;
+                sb.append(val.substring(beginIndex, endIndex));
+            }
+            if (mod > 0) {
+                sb.append(val.substring(div * MAX_STRING_LENGTH));
+            }
+            return sb.callToString();
+
+        } else {
+            return new ResultHandle("Ljava/lang/String;", this, val);
```
mkouba, line 332: I think that it would be worth mentioning this implementation details in the javadoc of this method, i.e. that a string < 65535 is stored in the constant pool while a string > 65535 is stored in the heap. A new string is constructed using the `StringBuilder` in the context of the current method body and the reference is returned.

geoand: @mkouba : https://github.com/quarkusio/gizmo/pull/152

gsmet:
1.6.1.Final is on its way to Maven Central. Let's include it for Quarkus 3.0.1.Final.

geoand:
Thanks

PR #152, opened 2023-04-12
Add note about usage of StringBuilder for loading strings

geoand:


src/main/java/io/quarkus/gizmo/BytecodeCreatorImpl.java
```diff
@@ -309,6 +309,11 @@ ResultHandle getOutgoingResultHandle() {
 
     }
 
+    /**
+     * Implementation detail: when the string to be loaded is larger than {@link BytecodeCreatorImpl#MAX_STRING_LENGTH},
+     * then a {@link StringBuilder} is used to construct the final String. This is done because UTF strings cannot be larger
```
Ladicek, line 314:
```suggestion
     * then a {@link StringBuilder} is used to construct the final String. This is done because strings constants cannot be larger
```

Ladicek (approved):

mkouba (approved):

PR #155, opened 2023-06-07
Update the usage docs - mention the switch constructs

mkouba:


Ladicek (approved):
I don't think many people will recognize the terms _rule block_ or _labeled statement group_, but they are technically correct per the JLS, so why not :-)

PR #157, opened 2023-07-14
Introduce ClassTransformer

mkouba:
- that can be used to add a method to an existing class

This can be really useful when transforming existing classes in Quarkus.

I'm thinking of a follow-up PR that would allow users to add bytecode to an existing method as well...

### DIFF
--- a/src/test/java/io/quarkus/gizmo/ClassTransformerTest.java
+++ b/src/test/java/io/quarkus/gizmo/ClassTransformerTest.java
@@ -1,0 +1,153 @@
+package io.quarkus.gizmo;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertThrows;
+import static org.junit.Assert.assertTrue;
+import static org.objectweb.asm.Opcodes.ACC_FINAL;
+import static org.objectweb.asm.Opcodes.ACC_PRIVATE;
+import static org.objectweb.asm.Opcodes.ACC_PUBLIC;
+import static org.objectweb.asm.Opcodes.ACC_SUPER;
+import static org.objectweb.asm.Opcodes.ACC_SYNTHETIC;
+
+import java.lang.reflect.Field;
+import java.lang.reflect.Method;
+import java.lang.reflect.Modifier;
+import java.util.function.Function;
+
+import org.junit.Test;
+import org.objectweb.asm.ClassReader;
+import org.objectweb.asm.ClassWriter;
+import org.objectweb.asm.Opcodes;
+
+public class ClassTransformerTest {
+
+    @Test
+    public void testTransfomer() throws Exception {
+        TestClassLoader cl = new TestClassLoader(getClass().getClassLoader());
+
+        String className = "com/MyTest";
+        String myInterfaceName = MyInterface.class.getName().replace('.', '/');
+        String objectName = "java/lang/Object";
+
+        ClassWriter cv = new ClassWriter(ClassWriter.COMPUTE_FRAMES | ClassWriter.COMPUTE_MAXS);
+        cv.visit(Opcodes.V11, ACC_PUBLIC | ACC_SUPER | ACC_SYNTHETIC, className, null, objectName,
+                new String[] { myInterfaceName });
+        cv.visitSource(null, null);
+
+        assertThrows(NullPointerException.class, () -> new ClassTransformer(className, null));
+        ClassTransformer classTransformer = new ClassTransformer(className, cv);
+
+        // private String latVal;
+        FieldCreator lastValue = classTransformer.addField(FieldDescriptor.of(className, "lastVal", String.class));
+        lastValue.setModifiers(ACC_PRIVATE);
+
+        // public String toStr;
+        FieldCreator toStr = classTransformer.addField("toStr", String.class);
+        toStr.setModifiers(ACC_PUBLIC);
+
+        // public MyTest() {
+        //    this.toStr = "Baf!";
+        // }
+        MethodCreator constructor = classTransformer.addMethod(MethodDescriptor.ofConstructor(className));
+        constructor.invokeSpecialMethod(MethodDescriptor.ofConstructor(Object.class), constructor.getThis());
+        constructor.writeInstanceField(toStr.getFieldDescriptor(), constructor.getThis(), constructor.load("Baf!"));
+        constructor.returnVoid();
+
+        // public final String transform(String val) {
+        //    if (lastVal != null && lastVal.equals("bar")) {
+        //       return "baZ";
+        //    }
+        //    lastVal = val;
+        //    if (val.equals("bar")) {
+        //       return val;
+        //    }
+        //    return val.toUpperCase();
+        // }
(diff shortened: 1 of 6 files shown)


Ladicek:
I think this is a great idea, but I don't think we should expose standalone access to `MethodCreator`. How about an alternative to `ClassCreator` called, say, `ClassModifier`, that would expose access to other Gizmo APIs (like `MethodCreator` or possibly `FieldCreator` later?). I'm thinking something like this: https://github.com/Ladicek/gizmo/commit/628ddf516a344a18f61d8f579668628c0031ff9d

mkouba:
> I think this is a great idea, but I don't think we should expose standalone access to `MethodCreator`. How about an alternative to `ClassCreator` called, say, `ClassModifier`, that would expose access to other Gizmo APIs (like `MethodCreator` or possibly `FieldCreator` later)?

Yes, that looks better. However, I don't think that `ClassModifier` is a good name because it reminds `Class.getModifiers()`... maybe `ClassTransformer`?

But I'm not quite sure we will be able to incorporate the "adding bytecode to an existing method" feature into this API...

Good point about the `FieldCreator`!

Ladicek:
`ClassTransformer` is a good name, agree.

Do you have any idea how you wanted the "add bytecode to existing method" thing to look like? Do you want to add instructions at the beginning and/or the end of the method, or something more general? What I think we could expose relatively easily is to _rename_ a method -- then you can create a new one that would replace the original and you call to it whenever you need. WDYT?

mkouba:
> `ClassTransformer` is a good name, agree.

:+1: 

> 
> Do you have any idea how you wanted the "add bytecode to existing method" thing to look like? Do you want to add instructions at the beginning and/or the end of the method, or something more general? What I think we could expose relatively easily is to _rename_ a method -- then you can create a new one that would replace the original and you call to it whenever you need. WDYT?

So I was thinking of adding some bytecode at the beginning/end of a method. For example if you need to initialize the fields in the constructor.

Renaming a method would be definitely a useful addition. The same applies to modyfing the modifiers ;-)

src/test/java/io/quarkus/gizmo/ClassTransformerTest.java
```diff
@@ -0,0 +1,105 @@
...
+        classTransformer.close();
+
+        cv.visitEnd();
+        cl.write(className, cv.toByteArray());
+
+        Class<?> clazz = cl.loadClass("com.MyTest");
+        assertTrue(clazz.isSynthetic());
+        MyInterface myInterface = (MyInterface) clazz.getDeclaredConstructor().newInstance();
+        Method transformMethod = clazz.getMethod("transform", String.class);
+        assertTrue(Modifier.isFinal(transformMethod.getModifiers()));
```
Ladicek, line 97: I would use try-with-resources for the `ClassTransformer` and use reflection here to also assert something about the field.

Ladicek (approved):
Otherwise LGTM.

src/test/java/io/quarkus/gizmo/ClassTransformerTest.java
mkouba: I don't think that we need to test the `AutoCloseable` functionality... Field assertion added.

mkouba:
> An interesting thing from API design perspective is that modifying existing stuff requires more careful handling of the `ClassVisitor`. Adding new things can be pretty much willy nilly (until `visitEnd()` is called), but modifying existing requires the `ClassTransformer` to be notified of all events that happen on the visitor.
> 
> So either the `ClassTransformer` will itself be a `ClassVisitor`, or it will expose one, something like that.

Hm, sounds like fun ;-). Anyway, let's add one thing at a time.

Ladicek:
Ah, there's one other thing worth discussing: both `FieldDescriptor` and `MethodDescriptor` require that the user knows the full [binary] name of the class they're transforming, even though that typically is not necessary (e.g. `FieldCreatorImpl` never uses that part of the `FieldDescriptor`, and I suspect `MethodCreatorImpl` is the same).

Maybe we should expose `addField` / `addMethod` methods that don't require the full descriptor?

mkouba:
> Ah, there's one other thing worth discussing: both `FieldDescriptor` and `MethodDescriptor` require that the user knows the full [binary] name of the class they're transforming, even though that typically is not necessary (e.g. `FieldCreatorImpl` never uses that part of the `FieldDescriptor`, and I suspect `MethodCreatorImpl` is the same).
> 
> Maybe we should expose `addField` / `addMethod` methods that don't require the full descriptor?

Hm, I used `FieldDescriptor` and `MethodDescriptor` because they declare a lot of useful factory methods, that also handle the conversion to internal class names, etc. And I'd like to avoid adding many `addMethod()`/`addField()` variants to the `ClassTransformer`.

Ladicek:
Mirroring what `ClassCreator` does would be fine IMHO. Perhaps even expected :shrug:

mkouba:
> Mirroring what `ClassCreator` does would be fine IMHO. Perhaps even expected shrug

The thing is that `ClassCreator` knows the `className` but the `ClassVisitor` does not, so we couldn't really use the `MethodDescriptor` internally :disappointed: .

Ladicek:
Just use `null` as the class name? Should work, and if there's a NPE somewhere, it's a bug.

We could also change `ClassTransformer` constructor to take the name of the transformed class.

mkouba:
> Just use `null` as the class name? Should work, and if there's a NPE somewhere, it's a bug.

I think that `DescriptorUtils.objectToInternalClassName()` would fail with IAE... why do you think it's a bug? I would say it's an incomplete descriptor definition instead ;-).

> We could also change `ClassTransformer` constructor to take the name of the transformed class.

That would work.

Ladicek:
Fair enough, it's probably better if we don't allow constructing incomplete descriptors :-)

mkouba:
I think that this PR is now ready! ;-)

Ladicek (approved):

Ladicek:
Hi @mkouba, I know I promised you to merge this PR, but then I started thinking and realized that the API we have now doesn't allow more complex transformations that we talked about above. I came up with an API that I believe is better (and allowed implementing method rename fairly easily as a proof): https://github.com/Ladicek/gizmo/commits/class-transformer Please take a look when you have time, and thanks for understanding :-)

mkouba:
> Hi @mkouba, I know I promised you to merge this PR, but then I started thinking and realized that the API we have now doesn't allow more complex transformations that we talked about above. I came up with an API that I believe is better (and allowed implementing method rename fairly easily as a proof): https://github.com/Ladicek/gizmo/commits/class-transformer Please take a look when you have time, and thanks for understanding :-)

It looks good. But I'm not so sure about the name of the `ClassTransformer#wrap(ClassVisitor delegate)` method. It's not very informative.

Ladicek:
Fair enough, WDYT about `wrapVisitor`? Or `apply[ToVisitor]`?

mkouba:
> Fair enough, WDYT about `wrapVisitor`? Or `apply[ToVisitor]`?

`apply()`/`applyTo()` sounds good.

Ladicek:
Superseded by #158.

PR #158, opened 2023-08-15
Introduce ClassTransformer

Ladicek:


Ladicek:
This supersedes #157, so I added all the original reviewers.

src/main/java/io/quarkus/gizmo/ClassTransformer.java
```diff
@@ -0,0 +1,196 @@
...
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Objects;
+
+import org.objectweb.asm.ClassVisitor;
+import org.objectweb.asm.MethodVisitor;
+
+/**
+ * Allows transforming an existing class by wrapping an ASM {@link ClassVisitor}.
```
mkouba, line 11: `Allows transforming ...` -> `Transforms ...` or `Allows for transformation of...`?

src/main/java/io/quarkus/gizmo/ClassTransformer.java
```diff
@@ -0,0 +1,196 @@
...
+ * of the original visitor.
+ * <p>
+ * In other words, this class acts like a builder. First, the class transformation
+ * must be described using {@code addMethod()}, {@code addField()}, etc. Then,
+ * a {@code ClassVisitor} must be created using {@code applyTo()}. Afterwards,
+ * the {@code ClassTransformer} instance should be discarded.
+ */
+public class ClassTransformer {
+
+    private static class NamedDescriptor {
```
mkouba, line 23: I would move this class down in the source file... it's private, not part of the API, so I see no reason to show it as the first thing when the `ClassTransformer` source is opened.

mkouba (approved):
Looks great! :+1:

src/main/java/io/quarkus/gizmo/ClassTransformer.java
Ladicek: I spent some time thinking about this exact sentence, because I didn't like it too much either. But "Transforms an existing class" is better, I'll fix that.

src/main/java/io/quarkus/gizmo/ClassTransformer.java
Ladicek: Fair enough, moving it to the end should be fine.

src/main/java/io/quarkus/gizmo/ClassTransformer.java
Ladicek: Done.

src/main/java/io/quarkus/gizmo/ClassTransformer.java
Ladicek: Done.

src/main/java/io/quarkus/gizmo/Switch.java
```diff
@@ -21,40 +21,36 @@ public interface Switch<T> {
      * <p>
      * For example, if fall through is disabled then:
      * 
-     * <pre>
-     * <code>
+     * <pre>{@code
```
mkouba, line 24: What's the benefit of `<pre>{@code`? Because readability it is not...

mkouba:
> FYI, I added one more commit that adds `ClassTransformer.addInterface()` and allows using `enumSwitch()`. It's an extra commit for easier review. I can drop the `addInterface()` stuff, I added that just to make testing easier, but I also thought it could be useful.

Nice!

src/main/java/io/quarkus/gizmo/Switch.java
Ladicek: It actually is way more readable in IntelliJ, but I confess I didn't try Eclipse :-)

I'll revert these changes, as they are not relevant to the present PR.

Ladicek: Done.

mkouba: Well, I was not talking about the rendered javadoc but pure text editor :shrimp:

Ladicek: Ah OK. I always prioritize the rendered form over the source text, but I agree the source is ugly. Unfortunately, `{@snippet ...}` was only added in JDK 18 and the [Codesnippet Doclet](https://github.com/jtulach/codesnippet4javadoc) is GPL3, so I'm hesitant to use it anywhere.

Ladicek:
Added one more commit that allows adding/removing modifiers from the class, its methods and fields.

PR #167, opened 2023-11-22
Simplify constructor MethodCreator

alesj:


src/main/java/io/quarkus/gizmo/ClassCreator.java
```diff
@@ -85,6 +85,14 @@ public ClassCreator(ClassOutput classOutput, String name, String signature, Stri
         this(null, classOutput, name, signature, superClass, ACC_PUBLIC | ACC_SUPER | ACC_SYNTHETIC, interfaces);
     }
 
+    public MethodCreator getConstructorAsMethodCreator(String... parameters) {
```
dmlloyd, line 88: Seems fine to me but I would not call the method this as it implies you are getting a constructor rather than a method builder. Maybe `getConstructorMethodBuilder` instead?

alesj: Renamed.

dmlloyd (approved):
LGTM

gsmet:
@mkouba @Ladicek could you have a look at this one?

src/main/java/io/quarkus/gizmo/ClassCreator.java
mkouba: Hm, I would personally use something like `getConstructorCreator()`. It's not ideal but aligns with thet current "wording". Also `MethodCreator` is not a real builder...

Ladicek: Agree, it should be called `getConstructorCreator()`.

Ladicek (approved):

mkouba (approved):

PR #172, opened 2024-01-24
Gizmo utils - add more convenient methods for list, set and map

mkouba:
- counterparts for List.of(), List.of(...), List.copyOf(Collection), Set.of(), Set.of(...), Set.copyOf(Collection), Map.of() and Map.copyOf(Map)

### DIFF
--- a/src/main/java/io/quarkus/gizmo/Gizmo.java
+++ b/src/main/java/io/quarkus/gizmo/Gizmo.java
@@ -74,7 +74,7 @@ public final class Gizmo {
 
     /**
      * <pre>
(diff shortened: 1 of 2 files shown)


mkouba:
There's a little bit more formatting changes that I would expect but I used the default Quarkus formatting rules so :shrug:

Ladicek (approved):

Ladicek:
LGTM, though the way how a `StringBuilder` is constructed in the test is sub-par, see `Gizmo.newStringBuilder()` :-)

Also, looks like we don't have a formatter set up for Gizmo yet, we should add that.

PR #173, opened 2024-01-24
Add support for primitive conversions

Ladicek:


mkouba (approved):
Very nice!

Ladicek:
Coincidence indeed -- I need this for my work on CDI method invokers in ArC, where method invokers automatically perform invocation conversions (that is, widening conversions), and I figured I could as well implement all primitive conversions (not much more code).

PR #178, opened 2024-02-05
Improve type conversions

### LINKED ISSUE #177: Smart `cast` operation
People are frequently tripping over casting, because primitive casts, object casts, and boxing/unboxing are all done in different ways. Provide a smart `cast` operation with the following support:

* Cast primitives to primtives as per `convertPrimitive`
* Cast objects to objects as per `checkCast`

And optionally:

* Cast primitives to primitive box objects by doing primitive-to-primitive conversion (if necessary) followed by `BoxType.valueOf(...)`
* Cast primitive box objects to primitives by doing `BoxType.xxxValue()` followed by primitive-to-primitive conversion (if necessary)
* Cast box objects to box objects by unboxing, converting, and then re-boxing

Since `checkCast` allows a `String` argument, primitive types and box types would need to be recognized by name.

Ladicek:
Fixes #177

### DIFF
--- a/USAGE.adoc
+++ b/USAGE.adoc
@@ -325,6 +325,24 @@ s.caseOf("baz", bc -> {
 s.defaultCase(bc -> bc.assign(ret, bc.loadNull()));
 ----
 
+=== Type conversions (casting)
+
+There are 3 methods for type conversions:
+
+* `checkCast()`
+* `convertPrimitive()`
+* `smartCast()`
+
+The `checkCast()` method emits the `checkcast` instruction, which performs a dynamic type check and a reference conversion.
+If the value that is supposed to be converted is statically known to be of a primitive type, the `checkCast()` method emits a boxing conversion first, before the type conversion.
+
+The `convertPrimitive()` method emits primitive conversion instructions (`i2l`, `d2i` etc.).
+If the value that is supposed to be converted is statically known to be of a primitive wrapper class, the `convertPrimitive()` method emits an unboxing conversion first, before the type conversion.
+
+The `smartCast()` method emits a sequence of instructions to cast the value to the given target type, using boxing, unboxing, primitive and reference conversions.
+For example, it allows casting `java.lang.Integer` to `java.lang.Long` by an unboxing conversion, followed by a primitive conversion, followed by a boxing conversion.
+This method is _not_ equivalent to the casting conversion described by Java Language Specification; it is a superset.
+
 == High-level Utilities
 
 The `Gizmo` class contains many utilities for generating common code sequences.
@@ -350,7 +368,7 @@ There's also a variant that takes a statically known initial capacity: `Gizmo.ne
 
 `Gizmo.newHashSet(BytecodeCreator target)` generates an invocation of `new HashSet()` into `target`.
 
-`Gizmo.newHashMap(BytecodeCreator target)` generates an invocation of `new HashMap()` into target.
+`Gizmo.newHashMap(BytecodeCreator target)` generates an invocation of `new HashMap()` into `target`.
 
 === StringBuilders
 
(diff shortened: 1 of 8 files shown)


Ladicek:
Probably best to review each commit in isolation.

I made the "smart cast" operation a `Gizmo.smartCast()` static method. It felt wrong to make it a method on `BytecodeCreator`, because it doesn't correspond to anything the JVM would ever do, but I realize this makes it less discoverable. I could probably be persuaded to move it.

mkouba:
> I made the "smart cast" operation a `Gizmo.smartCast()` static method. It felt wrong to make it a method on `BytecodeCreator`, because it doesn't correspond to anything the JVM or the Java language would do, but I realize this makes it less discoverable. I could probably be persuaded to move it.

We should at least add link/note to `Gizmo.smartCast()` in the javadoc of `BytecodeCreator#checkCast()` and `BytecodeCreator#convertPrimitive()`...

gsmet:
Yeah, I think we should favor discoverability.

mkouba (approved):

Ladicek:
OK, so I think I _finally_ fixed primitive conversions in case of `byte`, `short` and `char` (my previous belief that these conversions are implicit, since the values are `int`s on the JVM stack, was sorely mistaken).

So now we just need to agree on where the "smart cast" operation should be.

mkouba:
> So now we just need to agree on where the "smart cast" operation should be.

For me, the static method is fine but on the other hand I don't think that the `BytecodeCreator` must only contain stuff that would correspond to anything the JVM or the Java language would do... so maybe you should be persuaded to move it ;-).

gsmet:
If it wasn't clear, I'm in the `BytecodeCreator` camp :).

Ladicek:
OK, you guys have made your opinion very clear. Moved to `BytecodeCreator` :-)

src/main/java/io/quarkus/gizmo/BytecodeCreator.java
```diff
@@ -876,24 +893,107 @@ default ResultHandle checkCast(ResultHandle resultHandle, Class<?> castTarget) {
...
+     * to the target type. Other (reference) conversions are not checked statically.
+     * <p>
+     * This method is <em>not</em> equivalent to the casting conversion described by Java
+     * Language Specification; it is a superset.
+     *
+     * @param value the value to be casted
+     * @param castTarget the type to which the {@code value} should be casted
+     * @return the casted value
+     */
+    default ResultHandle smartCast(ResultHandle value, String castTarget) {
```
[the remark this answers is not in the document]
Ladicek: Fair point, I'll move it.

mkouba (approved):

USAGE.adoc
```diff
@@ -325,6 +325,24 @@ s.caseOf("baz", bc -> {
...
+
+The `checkCast()` method emits the `checkcast` instruction, which performs a dynamic type check and a reference conversion.
+If the value that is supposed to be converted is statically known to be of a primitive type, the `checkCast()` method emits a boxing conversion first, before the type conversion.
+
+The `convertPrimitive()` method emits primitive conversion instructions (`i2l`, `d2i` etc.).
+If the value that is supposed to be converted is statically known to be of a primitive wrapper class, the `convertPrimitive()` method emits an unboxing conversion first, before the type conversion.
+
+The `smartCast()` method emits a sequence of instructions to cast the value to the given target type, using boxing, unboxing, primitive and reference conversions.
+For example, it allows casting `java.lang.Integer` to `java.lang.Long` by an unboxing conversion, followed by a primitive conversion, followed by a boxing conversion.
+This method is _not_ equivalent to the casting conversion described by Java Language Specification; it is a superset.
```
dmlloyd, line 344: Is the new method a superset of the other two? If so could the other two be deprecated? If not could that fact be documented along with the specific differences?

Ladicek: It is a superset indeed.

I don't really see a reason for deprecating the other two methods. Or, I mean, I don't see anything obsolete about them. Their javadoc points to `smartCast()` as a "smarter" alternative, which I think should be good enough.

mkouba: -1 for deprecation. Sometimes it's not necessary to be that "smart" and the simple variants are good enough. And the javadoc is IMO clear.

holly-cummins:
It looks like this got lost a bit, so nudging it. :)

dmlloyd (approved):

Ladicek:
Ah sorry, totally forgot about this!

Issue #181, opened 2024-03-30
Loan pattern api for ClassCreator to control life cycle of object from API 


poldinik:
I would be nice if ClassCreator provide a static API to produce an instance of ClassCreator with try/catch with resource inside API, instead to delegate to a calling function the responsibility to implement try/catch

```
public static void withClassCreator(ClassCreator.Builder builder, Consumer<ClassCreator> consumer){
        try(ClassCreator classCreator = builder.build()){
            consumer.accept(classCreator);
        }
    } 
```

An example of usage

```
withClassCreator(
                ClassCreator.builder()
                        .classOutput(cl)
                        .className("com.MyTest")
                        .interfaces(MyInterface.class),
                creator -> {
                    MethodCreator method = creator.getMethodCreator("transform", String.class, String.class);
                    ResultHandle existingVal = method.load("complete");
                    TryBlock tb = method.tryBlock();
                    tb.addCatch(IllegalStateException.class).load(34);
                    tb.invokeStaticMethod(MethodDescriptor.ofMethod(ExceptionClass.class.getName(), "throwIllegalState", "V"));
                    method.returnValue(existingVal);
                });
```

PR #182, opened 2024-03-30
feat[class-creator]: Add loan pattern to incapsulate life cycle of ClassCreator inside API 

poldinik:
#181

### DIFF
--- a/src/main/java/io/quarkus/gizmo/ClassCreator.java
+++ b/src/main/java/io/quarkus/gizmo/ClassCreator.java
@@ -35,6 +35,8 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
 import java.util.concurrent.atomic.AtomicInteger;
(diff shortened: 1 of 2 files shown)


poldinik:
> I like the idea of this change (note that a few of the APIs I have added to this project in the past also use this pattern, and the new JDK classfile API does as well). But IMO we should apply it more uniformly, as opposed to haphazardly applying it to different places with no real overall plan.

How we can lead ahead this feature more systematically? Which step you see I need to follow to propose it and contribute? I mean something like roadmap o design specification before actual develop and avoid waste of time if stakeholders don't like implementation

dmlloyd:
This is a good question that I do not have an answer to. Maybe start with a high-level proposal that encompasses the existing API usages for method, class, and block creators, and go from there?

dmlloyd:
I don't know; I'm not familiar with the issue or its resolution.

PR #203, opened 2025-02-05
Various performance improvements

Postremus:
- save on unneeded isScopedWithin calls
- optimize findActiveResultHandles to not need an additional set to hold handles
- only query arc.debug system property once
- remove additional list construction from getInputResultHandles
- use constants for equality check to determine load constants instead

Related to https://github.com/quarkusio/quarkus/issues/45631

This change saves about 25ms in dev mode hot reload time for the reproducer in 45631 with quarkus 2.18.1

### DIFF
--- a/src/main/java/io/quarkus/gizmo/BytecodeCreatorImpl.java
+++ b/src/main/java/io/quarkus/gizmo/BytecodeCreatorImpl.java
@@ -1047,9 +1047,9 @@ class BytecodeCreatorImpl implements BytecodeCreator {
             }
             return;
         }
-        if (!isScopedWithin(handle.getOwner())) {
-            // throw new IllegalStateException("Wrong owner for ResultHandle " + handle);
-        }
+        //if (!isScopedWithin(handle.getOwner())) {
+        // throw new IllegalStateException("Wrong owner for ResultHandle " + handle);
+        //}
         if (handle.getResultType() != ResultHandle.ResultType.SINGLE_USE) {
             if (handle.getType().equals("S") || handle.getType().equals("Z") || handle.getType().equals("I")
                     || handle.getType().equals("B") || handle.getType().equals("C")) {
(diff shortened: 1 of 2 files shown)


dmlloyd (commented):
Some of these changes (like avoiding reading a system property every time we use an `Operation`) seem like a clear win, but some (like caching the value of e.g. `Integer.valueOf(0)` which is already cached by the JDK) seem less clear. Do you have a breakdown of what changed the performance curve, and why you think it did?

Postremus:
Hey @dmlloyd 
> but some (like caching the value of e.g. Integer.valueOf(0) which is already cached by the JDK) seem less clear.
I dropped that change. Did not seem to improve too much. Might have been a bit too eager there.

I mainly use the visualvm profiler, and the times reported by quarkus to determine if a change has any improvement.

Main parts are the changes in findActiveResultHandle, and getInputResultHandles.

For getInputResultHandles - yeah, one less array list created. Method is called 64074 times. less garbage, same readability. Other similar places are called way less (e.g. 27k, and 10k times). Did not seem worth it to me to make the code there a bit more complicated.

findActiveResultHandle - only called 19373 times. But the changes save on creation of another linkedhashset, at the same readability imo.

commenting out the called in loadResultHandle to isScopedWithin, saved 200k+ invocations of that method. (yes, loadresulthandle is called 275k times in the reproducer.)

to summarize, I believe the changes themself are small enough, that I would not bother to do them in a normal REST application, for example.
However, since a lot of the methods in gizmo are called 10k+ times, even a tiny change can have an impact.

gsmet:
I'm all for optimizing Gizmo (especially since every call sites will benefit from it) but I wonder if we should also check if we are not doing something very suboptimal in the bytecode generation on the Quarkus side.

dmlloyd (commented):
I think this looks OK now. But note for future reference that just because a method has a high call count according to an *instrumentation* based profile (as opposed to a sampling profile) does not mean it actually has an overhead. Many small methods (especially in `java.base`) may be inlined or even intrinsified, so that in practice, they are always called zero times.

But I think what is here now is a good and simple improvement with little risk, so I'm giving it the "LGTM" stamp.

Postremus:
Thanks for pointing that out regarding inlining.

Postremus:
Hey @gsmet any chance to get a new gizmo release with this improvement for Quarkus 3.21?

gsmet:
@Postremus sure, thanks for pinging me. I released 1.9.0 (there was a few new features in `main`) today and it should land into Quarkus tonight through Dependabot.

PR #221, opened 2025-03-11
Gizmo2: add convenient default methods to ClassOutput

mkouba:
- these methods accept FQCN instead of ClassDesc

### DIFF
--- a/src/main/java/io/quarkus/gizmo2/ClassOutput.java
+++ b/src/main/java/io/quarkus/gizmo2/ClassOutput.java
@@ -10,6 +10,18 @@ import io.quarkus.gizmo2.creator.InterfaceCreator;
  * A container for created classes with a specific output strategy.
  */
 public interface ClassOutput {
+    
+    /**
+     * Add a new class.
+     *
+     * @param name the fully qualified (dot-separated) binary class name
+     * @param builder the
+     * @return the descriptor
+     */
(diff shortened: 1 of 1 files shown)


dmlloyd (approved):
Looks like you found a spot where I messed up the docs. 🤦🏻‍♂️

LGTM otherwise.

src/main/java/io/quarkus/gizmo2/ClassOutput.java
```diff
@@ -10,6 +10,18 @@
...
 public interface ClassOutput {
+    
+    /**
+     * Add a new class.
+     *
+     * @param name the fully qualified (dot-separated) binary class name
+     * @param builder the
+     * @return the descriptor
+     */
+    default ClassDesc class_(String name, Consumer<ClassCreator> builder) {
```
gsmet, line 21: Sorry for asking out of the blue, but I was wondering if we should go with things like `createClass` instead of `class_`? Now maybe this is coming from the Class File API but it's not very pretty?

Note that this might be a completely stupid question as I haven't looked at the API in depth but I think it's probably now or never to ask this question.

mkouba: It's not stupid at all. We should think about a good strategy because it's not only `class` that is reserved and cannot be used "as is". I have to admit that I like the concise underscore suffix strategy but I agree that it does not feel very "Java"-like.

PR #223, opened 2025-03-11
Gizmo2: fix ClassCreator

mkouba:
- make the generated class "public synthetic" by default
- fix TypeCreator#_implements(Class<?>)

### DIFF
--- a/src/main/java/io/quarkus/gizmo2/creator/TypeCreator.java
+++ b/src/main/java/io/quarkus/gizmo2/creator/TypeCreator.java
@@ -2,16 +2,16 @@ package io.quarkus.gizmo2.creator;
 
 import java.lang.constant.ClassDesc;
 import java.util.Collection;
-import java.util.Set;
 import java.util.function.Consumer;
 
 import io.github.dmlloyd.classfile.Signature;
 import io.github.dmlloyd.classfile.extras.reflect.AccessFlag;
 import io.github.dmlloyd.classfile.extras.reflect.ClassFileFormatVersion;
 import io.quarkus.gizmo2.Annotatable;
-import io.quarkus.gizmo2.desc.MethodDesc;
 import io.quarkus.gizmo2.StaticFieldVar;
+import io.quarkus.gizmo2.desc.MethodDesc;
 import io.quarkus.gizmo2.impl.TypeCreatorImpl;
+import io.quarkus.gizmo2.impl.Util;
 
 /**
  * A creator for a type.
@@ -68,6 +68,16 @@ public sealed interface TypeCreator extends Annotatable permits ClassCreator, In
     }
 
     /**
+     * Add the {@code public} access flag to the type.
+     */
+    void public_();
+
+    /**
+     * Remove the {@code public} access flag.
+     */
+    void packagePrivate();
+
+    /**
      * Set the source file name for this type.
      *
      * @param name the source file name (must not be {@code null})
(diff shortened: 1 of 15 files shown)


src/main/java/io/quarkus/gizmo2/impl/ClassCreatorImpl.java
```diff
@@ -19,6 +19,7 @@
 public final class ClassCreatorImpl extends TypeCreatorImpl implements ClassCreator {
     public ClassCreatorImpl(final ClassDesc type, final ClassBuilder zb) {
         super(type, zb, 0);
+        withFlags(AccessFlag.PUBLIC, AccessFlag.SYNTHETIC);
```
dmlloyd, line 22: I guess we'll also need a `withoutFlags` then to override the default. WDYT?

src/main/java/io/quarkus/gizmo2/creator/TypeCreator.java
```diff
@@ -102,6 +101,7 @@ default void implements_(Class<?> interface_) {
         if (! interface_.isInterface()) {
             throw new IllegalArgumentException("Only interfaces may be implemented");
         }
+        implements_(ClassDesc.of(interface_.getName()));
```
dmlloyd, line 104: Heh, oops!

src/main/java/io/quarkus/gizmo2/impl/ClassCreatorImpl.java
dmlloyd: Oh, here's an idea, what about methods like `cc.public_();` `cc.packagePrivate()` etc.? They could intelligently set/clear flags as needed.

src/main/java/io/quarkus/gizmo2/creator/TypeCreator.java
dmlloyd: For cases like this you should use `Util.classDesc(interface_)` btw. It caches the descriptor for each type.

src/main/java/io/quarkus/gizmo2/impl/InterfaceCreatorImpl.java
```diff
@@ -48,6 +52,8 @@ public MethodDesc method(final String name, final Consumer<AbstractMethodCreator
     }
 
     void accept(final Consumer<InterfaceCreator> builder) {
+        preAccept();
```
mkouba, line 55: `preAccept()` and `postAccept()` must be called in order to propagate flags etc.

src/main/java/io/quarkus/gizmo2/impl/StaticFieldCreatorImpl.java
```diff
@@ -17,7 +18,7 @@ public final class StaticFieldCreatorImpl extends FieldCreatorImpl implements St
     private Consumer<BlockCreator> initializer;
 
     public StaticFieldCreatorImpl(final TypeCreatorImpl tc, final ClassDesc owner, final String name) {
-        super(owner, name, tc);
+        super(owner, name, tc, AccessFlag.STATIC.mask());
```
mkouba, line 21: Just make static fields static ;-)

src/main/java/io/quarkus/gizmo2/impl/FieldCreatorImpl.java
```diff
@@ -11,19 +11,21 @@
...
 
-public abstract sealed class FieldCreatorImpl extends AnnotatableCreatorImpl implements FieldCreator permits StaticFieldCreatorImpl, InstanceFieldCreatorImpl {
+public abstract sealed class FieldCreatorImpl extends AnnotatableCreatorImpl implements FieldCreator
+        permits StaticFieldCreatorImpl, InstanceFieldCreatorImpl {
     protected final ClassDesc owner;
     protected final String name;
     protected final TypeCreatorImpl tc;
-    Signature.ClassTypeSig genericType = Signature.ClassTypeSig.of(CD_int);
-    ClassDesc type = CD_int;
+    Signature.ClassTypeSig genericType = Signature.ClassTypeSig.of(CD_Integer);
```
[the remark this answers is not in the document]
mkouba: Uf, I don't remember but I'll add some tests for this soon so we'll run into this problem again ;-)

PR #246, opened 2025-03-17
Gizmo2: introduce ListOps#get()

mkouba:
- only print the class if the "printClass" system property is set

### DIFF
--- a/src/main/java/io/quarkus/gizmo2/creator/ops/ListOps.java
+++ b/src/main/java/io/quarkus/gizmo2/creator/ops/ListOps.java
@@ -2,6 +2,7 @@ package io.quarkus.gizmo2.creator.ops;
 
 import java.util.List;
 
+import io.quarkus.gizmo2.Constant;
 import io.quarkus.gizmo2.Expr;
 import io.quarkus.gizmo2.creator.BlockCreator;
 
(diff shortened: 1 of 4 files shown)


src/main/java/io/quarkus/gizmo2/creator/ops/ListOps.java
```diff
@@ -29,4 +29,14 @@ public ListOps(final BlockCreator bc, final Expr obj) {
...
         super(receiverType.asSubclass(List.class), bc, obj);
     }
+    
+    /**
+     * Generate a call to {@link List#get(int)}.
+     *
+     * @param index the index of the element to return (must not be {@code null})
+     * @return the expression of the result (not {@code null})
+     */
+    public Expr get(Expr index) {
```
dmlloyd, line 39: I would add an overload which accepts an `int`-typed index for this.

dmlloyd (approved):
LGTM, one minor suggestion though (see comment).

src/main/java/io/quarkus/gizmo2/creator/ops/ListOps.java
mkouba: Done.

Issue #244, opened 2025-03-17
Gizmo2: switch statement results in IllegalStateException


mkouba:
A [simple test](https://github.com/mkouba/gizmo/blob/switch-test/src/test/java/io/quarkus/gizmo2/SwitchTest.java) results in:
```
java.lang.IllegalStateException: This block is not active
	at io.quarkus.gizmo2/io.quarkus.gizmo2.impl.BlockCreatorImpl.checkActive(BlockCreatorImpl.java:1174)
	at io.quarkus.gizmo2/io.quarkus.gizmo2.impl.BlockCreatorImpl.replaceLastItem(BlockCreatorImpl.java:1155)
	at io.quarkus.gizmo2/io.quarkus.gizmo2.impl.BlockCreatorImpl.return_(BlockCreatorImpl.java:1009)
	at io.quarkus.gizmo2/io.quarkus.gizmo2.SwitchTest.lambda$2(SwitchTest.java:50)
	at io.quarkus.gizmo2/io.quarkus.gizmo2.impl.BlockCreatorImpl.accept(BlockCreatorImpl.java:789)
	at io.quarkus.gizmo2/io.quarkus.gizmo2.impl.ExecutableCreatorImpl.doCode(ExecutableCreatorImpl.java:166)
	at io.quarkus.gizmo2/io.quarkus.gizmo2.impl.ExecutableCreatorImpl.lambda$8(ExecutableCreatorImpl.java:152)
```

Issue #237, opened 2025-03-13
Lock creator states


dmlloyd:
Most creators need to have a state progression:

* Body not created yet (declare variables, parameters, superclasses, etc.)
* Body creation in progress (add code and/or members)
* Body complete (final state, all values are locked in)

Right now, it's possible for some creators to modify things in later states that shouldn't be modified in those states. Solve this as uniformly as possible.

PR #247, opened 2025-03-18
Complete rework of `switch`

### LINKED ISSUE #228: Gizmo2: Switch expressions
Never quite got around to finishing them.

dmlloyd:
Fixes #228, #244, part of #237


src/main/java/io/quarkus/gizmo2/creator/BlockCreator.java
```diff
@@ -1,6 +1,6 @@
 package io.quarkus.gizmo2.creator;
 
-import static java.lang.constant.ConstantDescs.CD_boolean;
+import static java.lang.constant.ConstantDescs.*;
```
mkouba, line 3: I think that we should not use the `*` imports. In fact, gizmo should probably use the same [rules as Quarkus](https://github.com/quarkusio/quarkus/blob/main/independent-projects/ide-config/pom.xml).

dmlloyd: I agree for class imports. For static imports, especially for classes like this which are designed to be used in this way, I disagree. It seems pointless to expand them all out in this case.

PR #249, opened 2025-03-19
Gizmo2: fix while loop and make it possible to continue a foreach loop

mkouba:
- also add a few more tests for loops


dmlloyd (approved):
Oops! Good catch.

PR #261, opened 2025-03-26
Gizmo2: fix BlockCreator#unbox()

mkouba:
I'm not quite sure how to work with primitive types that cannot be represented directly in the constant pool (`byte`, `short`, `char`). So the test is not complete.

NOTE: In Gizmo1, we have some autoboxing functionality implemented. It might make sense to consider something similar in Gizmo2. @Ladicek knows probably more...

### DIFF
--- a/src/test/java/io/quarkus/gizmo2/BoxUnboxTest.java
+++ b/src/test/java/io/quarkus/gizmo2/BoxUnboxTest.java
@@ -1,0 +1,100 @@
+package io.quarkus.gizmo2;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import java.util.List;
+import java.util.function.Supplier;
+
+import org.junit.jupiter.api.Test;
+
+public class BoxUnboxTest {
+
+    @Test
+    public void testBox() {
+        TestClassMaker tcm = new TestClassMaker();
+        Gizmo g = Gizmo.create(tcm);
+        g.class_("io.quarkus.gizmo2.Box", cc -> {
+            cc.staticMethod("test", mc -> {
+                // static List test() {
+                //    Boolean boolVal = true;
+                //    Integer intVal = 65536;
+                //    Long longVal = 0x7fffffffffffffffL;
+                //    Float floatVal = 1.1;
+                //    Double doubleVal = 1.2;
+                //    return List.of(boolVal, byteVal, charVal, shortVal, intVal, longVal, floatVal, doubleVal);
+                // }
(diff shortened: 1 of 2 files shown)


Ladicek:
Yeah there's a couple things that you have to do explicitly in current state of Gizmo 2 (boxing/unboxing, but also capturing variables in lambdas/anonymous classes) that were implicit in Gizmo 1. Not sure how I feel about that.

src/test/java/io/quarkus/gizmo2/BoxUnboxTest.java
```diff
@@ -0,0 +1,100 @@
...
+                // }
+                var b = mc.parameter("b", Boolean.class);
+                var i = mc.parameter("i", Integer.class);
+                var l = mc.parameter("l", Long.class);
+                var f = mc.parameter("f", Float.class);
+                var d = mc.parameter("d", Double.class);
+                mc.returning(int.class);
+                mc.body(bc -> {
+                    // WORKAROUND: we need to use local vars for types where unboxing involves cmp/cmpg
+                    // TODO: file a new issue
```
Ladicek, line 78: +1

src/test/java/io/quarkus/gizmo2/BoxUnboxTest.java
```diff
@@ -0,0 +1,100 @@
...
+                    var lu = bc.define("lv", bc.unbox(l));
+                    var fu = bc.define("fv", bc.unbox(f));
+                    var du = bc.define("dv", bc.unbox(d));
+                    bc.unless(bc.unbox(b), fail -> fail.return_(1));
+                    bc.if_(bc.ne(bc.unbox(i), 10), fail -> fail.return_(2));
+                    bc.if_(bc.ne(lu, 100l), fail -> fail.return_(3));
+                    bc.if_(bc.ne(fu, (float) 1.2), fail -> fail.return_(4));
+                    bc.if_(bc.ne(du, 2.1), fail -> fail.return_(5));
```
Ladicek, line 86: I would expect unboxing on these lines too?

dmlloyd:
> Yeah there's a couple things that you have to do explicitly in current state of Gizmo 2 (boxing/unboxing, but also capturing variables in lambdas/anonymous classes) that were implicit in Gizmo 1. Not sure how I feel about that.

We can definitely improve some of that. I just went for the shortest path to completion in a lot of cases. I think we could probably implement decent autoboxing/unboxing though.

dmlloyd (approved):

PR #263, opened 2025-03-27
inform Maven directly that Gizmo 2 is a Java 17 project

Ladicek:
This lets IntelliJ IDEA import the project with the correct language level, which allows running tests. Alas, IntelliJ IDEA isn't able to read the profile activated by the `build-release-17` file properly, so we need to duplicate the info.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -31,6 +31,15 @@
     </licenses>
 
     <properties>
+        <!--
+          - These properties should not be necessary due to the `build-release-17` control file being present,
+          - but IntelliJ does not import the project correctly without these, regardless, so set them directly
+          - to give it the clues it is sorely lacking.
+          -->
+        <maven.compiler.source>17</maven.compiler.source>
+        <maven.compiler.target>17</maven.compiler.target>
+        <maven.compiler.release>17</maven.compiler.release>
+
         <version.classfile>24.0</version.classfile>
         <version.junit>5.12.1</version.junit>
 


dmlloyd:
The control file `build-release-17` should already be doing this through the parent POM. If importing is going wrong, it'll be a parent bug IMO.

Ladicek:
That's funny. I tried enabling the profile explicitly in IDEA (because it wasn't enabled automatically), and I tried adding `<source>` / `<target>` into the profile, but I still got language level 11 for the `gizmo2.test` module :-/

dmlloyd:
I wonder if it is due to the configuration for the compiler plugin being directed at the specific execution for the base layer, instead of being the default.  Maybe IntelliJ can't figure out what's going on in that case.  I'll test that theory...

Ladicek:
I guess everything is possible :-) IntelliJ isn't even fine with just `maven.compiler.release`, I have to add `source` and `target` as well (maybe one of them would be enough, I didn't try).

pom.xml
```diff
@@ -31,6 +31,10 @@
     </licenses>
 
     <properties>
+        <maven.compiler.source>17</maven.compiler.source>
+        <maven.compiler.target>17</maven.compiler.target>
+        <maven.compiler.release>17</maven.compiler.release>
```
dmlloyd, line 36:
```suggestion
        <!-- These properties should not be necessary due to the `build-release-17` control file being present; but IntelliJ does not import the project correctly without these, regardless, so set them directly to give it the clues it is sorely lacking -->
        <maven.compiler.source>17</maven.compiler.source>
        <maven.compiler.target>17</maven.compiler.target>
        <maven.compiler.release>17</maven.compiler.release>
```

mkouba (approved):
Hm, who cares about IntelliJ IDEA? :wink: 

In any case, this change is ok.

pom.xml
mkouba: :-1: for this comment. AFAIK the `build-release-17` file might not be enough.

dmlloyd:
Yeah no matter what I change in the parent, it simply won't accept the test source code level from the profile configuration. I guess it's an IntelliJ bug. I'm fine with this change as long as you add a note like the one I suggested.

pom.xml
mkouba: I.e. `maven.compiler.source` is the standard for Maven projects.

Ladicek: Done, thanks!

dmlloyd (approved):

PR #264, opened 2025-03-27
fix primitive constants

Ladicek:
We need constants of types `byte`, `short` and `char` because without them, we cannot statically make a difference between them and `int`, which may lead to all kinds of issues. For example, we could emit an invocation of a different overload of a method than expected, if the target class has an overload of the method for both `byte` (or `short` or `char`) and `int`.

In addition to that, this commit also adds `Constant.of(Boolean)`, because `java.lang.Boolean` constants were not of the primitive type, unlike all others. This might be a bad idea, but all other constants of wrapper types are in fact constants of the corresponding primitive type, and consistency is warranted. We might later want to rethink that.

### DIFF
--- a/src/main/java/io/quarkus/gizmo2/Constant.java
+++ b/src/main/java/io/quarkus/gizmo2/Constant.java
@@ -99,6 +99,54 @@ public sealed interface Constant extends Expr, Constable permits ConstantImpl {
      * {@return a constant for the given value}
      * @param value the value to create a constant from (must not be {@code null})
      */
+    static Constant of(Byte value) {
+        return ConstantImpl.of(value);
+    }
+
+    /**
+     * {@return a constant for the given value}
+     * @param value the value to create a constant from
+     */
+    static Constant of(byte value) {
+        return ConstantImpl.of(value);
+    }
+
+    /**
+     * {@return a constant for the given value}
+     * @param value the value to create a constant from (must not be {@code null})
+     */
+    static Constant of(Short value) {
+        return ConstantImpl.of(value);
+    }
+
+    /**
+     * {@return a constant for the given value}
+     * @param value the value to create a constant from
+     */
+    static Constant of(short value) {
+        return ConstantImpl.of(value);
+    }
+
+    /**
+     * {@return a constant for the given value}
+     * @param value the value to create a constant from (must not be {@code null})
+     */
+    static Constant of(Character value) {
+        return ConstantImpl.of(value);
+    }
+
+    /**
+     * {@return a constant for the given value}
+     * @param value the value to create a constant from
+     */
+    static Constant of(char value) {
+        return ConstantImpl.of(value);
+    }
+
+    /**
+     * {@return a constant for the given value}
+     * @param value the value to create a constant from (must not be {@code null})
+     */
     static Constant of(Integer value) {
         return ConstantImpl.of(value);
     }
@@ -156,6 +204,14 @@ public sealed interface Constant extends Expr, Constable permits ConstantImpl {
      * @param value the value to create a constant from
      */
     static Constant of(double value) {
+        return ConstantImpl.of(value);
+    }
+
+    /**
+     * {@return a constant for the given value}
+     * @param value the value to create a constant from
+     */
+    static Constant of(Boolean value) {
         return ConstantImpl.of(value);
     }
 
--- a/src/main/java/io/quarkus/gizmo2/impl/GizmoImpl.java
+++ b/src/main/java/io/quarkus/gizmo2/impl/GizmoImpl.java
@@ -18,6 +18,8 @@ import io.quarkus.gizmo2.StaticFieldVar;
 import io.quarkus.gizmo2.creator.ClassCreator;
 import io.quarkus.gizmo2.creator.InterfaceCreator;
 import io.quarkus.gizmo2.impl.constant.ArrayVarHandleConstant;
+import io.quarkus.gizmo2.impl.constant.ByteConstant;
+import io.quarkus.gizmo2.impl.constant.CharConstant;
 import io.quarkus.gizmo2.impl.constant.ClassConstant;
 import io.quarkus.gizmo2.impl.constant.ConstantImpl;
 import io.quarkus.gizmo2.impl.constant.DoubleConstant;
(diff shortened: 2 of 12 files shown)


dmlloyd:
TL;DR I agree with the basic idea. There are some details to work out though.

Longer explanation & background:

As you guessed, I left off these types initially because they are all represented as `int` when loaded into bytecode, and I hoped to avoid the complexity which comes with these values. But you're right, we need a way to represent constants with different static types for easy method resolution at least (the user could always give an explicit signature and use int constants, but that's pretty unwieldy).

What complexity you may ask? Well...

Relying on `Constable` means that any time a constant of this type shows up in bytecode, we're doing a `ldc` with a dynamic constant whose initializer is basically taking an `int` constant and casting it to `byte` or `short` or whatever.  However, ideally when we generate code for a constant `byte` (for example) we'd want to just load the `int` constant and use it directly (since we generate the constant, we will know it's in range). This will prevent our constant table from getting too bloated in the common case.

So, to do this right, we need to do a couple of things:

* Create `ByteConstant`, `ShortConstant`, and `CharConstant` classes whose `writeCode` method is the same as `IntConstant` (maybe extract a common superclass called `IntLoadableConstant` or something)
  * `BooleanConstant` is already basically doing this correctly
* Modify `io.quarkus.gizmo2.impl.constant.ConstantImpl#of(java.lang.constant.Constable)` to detect `Byte`, `Short`, and `Character` (and `Boolean`, oops I forgot that one!), and call a type-specific `of` method much like `io.quarkus.gizmo2.impl.constant.ConstantImpl#of(java.lang.constant.ConstantDesc)` does when it gets an `Integer` or `Long` etc.
* Now the ugly one: modify `io.quarkus.gizmo2.impl.constant.ConstantImpl#of(java.lang.constant.DynamicConstantDesc<?>)` to detect a `DynamicConstantDesc` which matches the following described constants from the wrapper classes, and redirects them to the correct type-specific `of` method:
  * For these three, check the component `constantName()`, `constantType()`, and `bootstrapMethod()`:
    * `DynamicConstantDesc.ofNamed(BSM_EXPLICIT_CAST, DEFAULT_NAME, CD_byte, (int) value)`
    * `DynamicConstantDesc.ofNamed(BSM_EXPLICIT_CAST, DEFAULT_NAME, CD_char, (int) value)`
    * `DynamicConstantDesc.ofNamed(BSM_EXPLICIT_CAST, DEFAULT_NAME, CD_short, (int) value)`
  * For these two, just use `equals`:
    * `java.lang.constant.ConstantDescs#TRUE`
    * `java.lang.constant.ConstantDescs#FALSE`

This way, no matter what weird way the user sends in a Byte, Short, etc., we always get the same correctly-typed constant out of it.

I hope this makes sense!

Ladicek:
Yeah, I originally started doing something like that, but @mkouba's comment that `Byte`, `Short` and `Character` are `Constable made me lazy :-) I'll fix this properly, thanks!

Ladicek:
Done!

src/main/java/io/quarkus/gizmo2/impl/constant/ConstantImpl.java
```diff
@@ -72,8 +81,6 @@ public static ConstantImpl of(DynamicConstantDesc<?> dcd) {
             return of(desc);
         } else if (dcd instanceof VarHandle.VarHandleDesc desc) {
             return of(desc);
-        } else if (dcd.bootstrapMethod().equals(ConstantDescs.BSM_NULL_CONSTANT)) {
-            return ofNull(dcd.constantType());
```
dmlloyd, line 76: What happened here?

src/main/java/io/quarkus/gizmo2/Constant.java
```diff
@@ -95,6 +95,54 @@ static Constant of(Enum.EnumDesc<?> value) {
...
+    static Constant of(Byte value) {
+        return ConstantImpl.of(value);
+    }
+
+    /**
+     * {@return a constant for the given value}
+     * @param value the value to create a constant from
+     */
+    static Constant of(byte value) {
+        return ConstantImpl.of((Byte) value);
```
dmlloyd, line 111: These methods could just be `ConstantImpl.of(value)` right?

dmlloyd (commented):
Looks good other than two minor questions. Thanks!

src/main/java/io/quarkus/gizmo2/impl/GizmoImpl.java
```diff
@@ -98,6 +101,18 @@ public StringConstant stringConstant(String value) {
         return (StringConstant) constants.computeIfAbsent(value, StringConstant::new);
     }
 
+    public ByteConstant byteConstant(Byte value) {
+        return (ByteConstant) funnyHashCodeConstants.computeIfAbsent(new ByteConstant(value), Function.identity());
```
mkouba, line 105: Shouldn't we try to find a better name for `funnyHashCodeConstants`? :thinking:

mkouba (commented):
We could also extend the `BoxUnboxTest` now...

src/main/java/io/quarkus/gizmo2/impl/constant/ConstantImpl.java
Ladicek: This is duplicated a few lines below, with the comment `"wrong spelling" of null constant`. I figured one case is enough :-)

src/main/java/io/quarkus/gizmo2/Constant.java
Ladicek: Good point, will fix.

src/main/java/io/quarkus/gizmo2/impl/GizmoImpl.java
Ladicek: I would personally merge `constants` and `funnyHashCodeConstants` together (using the `funnyHashCodeConstants` approach), not sure why they're separate (except perhaps a little less allocations in the `constants` case?).

Ladicek:
> We could also extend the `BoxUnboxTest` now...

Let me take a look.

src/main/java/io/quarkus/gizmo2/Constant.java
Ladicek: Done.

mkouba (approved):

src/main/java/io/quarkus/gizmo2/impl/constant/ConstantImpl.java
```diff
@@ -35,8 +35,19 @@ public static StringConstant of(final String value) {
...
+        } else if (constable instanceof Boolean val) {
+            return of(val);
+/*
+        } else if (constable instanceof Byte val) {
+            return of(val);
+        } else if (constable instanceof Short val) {
+            return of(val);
+        } else if (constable instanceof Character val) {
+            return of(val);
+*/
```
dmlloyd, line 50: What happened here? Was this a test?

Ladicek: Ouch! Indeed this was a test. Sorry about that, fixed.

dmlloyd:
Awesome, thanks!

PR #268, opened 2025-03-28
Add convenient TypeCreator#constantField(String, Constant)

mkouba:


Ladicek (approved):
Seems reasonable to me.

dmlloyd (approved):

PR #271, opened 2025-03-28
fix the ArrayStore item

Ladicek:
The `ArrayStore` item used to emit an array store instruction based on the type kind of the array, which is always reference. It needs to do that based on the type kind of the _component type_ of the array.

Further, this commit replaces all occurences of "element type" with "component type", because that's what they really are. This doesn't make a difference with single-dimensional arrays, which is the vast majority of arrays, but it does make a difference with multi-dimensional arrays, where component type is the type of the inner array, while the element type is never an array type.

Finally, some tests for arrays are added.

### DIFF
--- a/src/main/java/io/quarkus/gizmo2/impl/ArrayStore.java
+++ b/src/main/java/io/quarkus/gizmo2/impl/ArrayStore.java
@@ -1,18 +1,22 @@
 package io.quarkus.gizmo2.impl;
 
+import java.lang.constant.ClassDesc;
 import java.util.function.BiFunction;
 
 import io.github.dmlloyd.classfile.CodeBuilder;
+import io.github.dmlloyd.classfile.TypeKind;
 
 final class ArrayStore extends Item {
     private final Item arrayExpr;
     private final Item index;
     private final Item value;
+    private final ClassDesc componentType;
 
-    ArrayStore(final Item arrayExpr, final Item index, final Item value) {
+    ArrayStore(final Item arrayExpr, final Item index, final Item value, final ClassDesc componentType) {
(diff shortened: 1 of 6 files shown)


src/main/java/io/quarkus/gizmo2/impl/ArrayStore.java
```diff
@@ -32,6 +36,6 @@ protected Node forEachDependency(final Node node, final BiFunction<Item, Node, N
     }
 
     public void writeCode(final CodeBuilder cb, final BlockCreatorImpl block) {
-        cb.arrayStore(arrayExpr.typeKind());
+        cb.arrayStore(TypeKind.from(componentType));
```
Ladicek, line 39: This is the actual bugfix. The rest of this commit is cleanup :-)

dmlloyd (approved):
Makes sense, thanks.

mkouba (approved):

mkouba:
@Ladicek @dmlloyd Speaking of arrays I find the default `io.quarkus.gizmo2.Expr.elem(Constant)`, `io.quarkus.gizmo2.Expr.elem(ConstantDesc)` and `io.quarkus.gizmo2.Expr.elem(Constable)` methods quite useless. Note that you can't do something like `expr.elem(Integer.valueOf(1))`  because of ambiguity. WDYT?

dmlloyd:
I would add an explicit `Integer` overload too, personally, otherwise we'd end up with a useless unbox-rebox cycle.

mkouba:
Ok, so `int`, `Integer` and `Expr` it is. I will send a PR.

PR #272, opened 2025-03-28
add StringBuilderOps

### LINKED ISSUE #257: Gizmo2: port StringBuilderGenerator
See https://github.com/quarkusio/gizmo/blob/main/src/main/java/io/quarkus/gizmo/Gizmo.java#L1097

Ladicek:
This is an equivalent of `StringBuilderGenerator` from Gizmo 1.x.

This commit also moves tests for existing `*Ops` classes to the `ops` package, similarly to `StringBuilderOpsTest`.

Resolves #257

### DIFF
--- a/src/main/java/io/quarkus/gizmo2/creator/ops/StringBuilderOps.java
+++ b/src/main/java/io/quarkus/gizmo2/creator/ops/StringBuilderOps.java
@@ -1,0 +1,98 @@
+package io.quarkus.gizmo2.creator.ops;
+
+import io.quarkus.gizmo2.Constant;
+import io.quarkus.gizmo2.Expr;
+import io.quarkus.gizmo2.creator.BlockCreator;
+
+/**
+ * Operations on {@link StringBuilder}. The expected usage pattern is:
+ * <ol>
+ * <li>Create an instance using {@link BlockCreator#withNewStringBuilder()}</li>
+ * <li>Append to it using {@link #append(Expr)}</li>
+ * <li>Create the final string using {@link #objToString()}</li>
+ * </ol>
+ * If you need to perform other operations on the {@code StringBuilder}
+ * that this class doesn't provide, you should create an instance
+ * using {@link BlockCreator#withStringBuilder(Expr)}, which allows
+ * you to pass an already created {@code StringBuilder}. This class
+ * itself doesn't provide access to the underlying object.
+ */
+public final class StringBuilderOps extends ObjectOps implements ComparableOps {
+
+    /**
+     * Construct a new instance on an existing {@link StringBuilder}.
+     *
+     * @param bc the block creator (must not be {@code null})
+     * @param obj the receiver string builder (must not be {@code null})
+     */
+    public StringBuilderOps(final BlockCreator bc, final Expr obj) {
+        super(StringBuilder.class, bc, obj);
+    }
+
+    /**
+     * Construct a new instance on a newly created {@link StringBuilder}.
+     *
+     * @param bc the block creator (must not be {@code null})
+     */
+    public StringBuilderOps(final BlockCreator bc) {
+        super(StringBuilder.class, bc, bc.define("stringBuilder", bc.new_(StringBuilder.class)));
+    }
+
+    /**
+     * Construct a new instance on a newly created {@link StringBuilder}
+     * with given {@code capacity}.
+     *
+     * @param bc the block creator (must not be {@code null})
+     * @param capacity the capacity of the newly created {@link StringBuilder}
+     */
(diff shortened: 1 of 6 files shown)


src/main/java/io/quarkus/gizmo2/creator/ops/StringBuilderOps.java
```diff
@@ -0,0 +1,82 @@
...
+     * may be obtained using {@link #getInstance()}.
+     *
+     * @param bc the block creator (must not be {@code null})
+     * @param capacity the capacity of the newly created {@link StringBuilder}
+     */
+    public StringBuilderOps(final BlockCreator bc, final int capacity) {
+        super(StringBuilder.class, bc, bc.new_(StringBuilder.class, Constant.of(capacity)));
+    }
+
+    public StringBuilderOps append(final Expr expr) {
```
dmlloyd, line 51: The rest of these methods should have javadoc as well.

src/main/java/io/quarkus/gizmo2/creator/ops/StringBuilderOps.java
```diff
@@ -0,0 +1,82 @@
...
+
+    public StringBuilderOps append(final char constant) {
+        return append(Constant.of(constant));
+    }
+
+    public StringBuilderOps append(final String constant) {
+        return append(Constant.of(constant));
+    }
+
+    public Expr callToString() {
```
mkouba, line 70: There's already `io.quarkus.gizmo2.creator.ops.ObjectOps.objToString()`...

Ladicek: Ah, good point! Indeed this can be reused. Will fix.

Ladicek: Done.

src/main/java/io/quarkus/gizmo2/creator/ops/StringBuilderOps.java
```diff
@@ -0,0 +1,83 @@
...
+
+    public StringBuilderOps append(final char constant) {
+        return append(Constant.of(constant));
+    }
+
+    public StringBuilderOps append(final String constant) {
+        return append(Constant.of(constant));
+    }
+
+    public Expr getInstance() {
```
dmlloyd, line 75: In this project (and pretty much everything since records were added to the language) I've been using the form `instance()` rather than `getInstance()`.

mkouba, line 75: I wonder if we shouldn't just make the `ObjectOps#obj` field public instead... :thinking:

mkouba (approved):
Awesome!

src/main/java/io/quarkus/gizmo2/creator/ops/StringBuilderOps.java
Ladicek: Ah yeah, good point. Will add.

src/main/java/io/quarkus/gizmo2/creator/ops/StringBuilderOps.java
Ladicek: Makes sense, will do.

mkouba: > I'm not a huge fan of public fields; at most it should be an `instance()` method on the base.

Me neither, except for `public final` fields.

@Ladicek why do we need this method for `StringBuilderOps` anyway?

Ladicek: I don't actually think we need it, because there's a way to create `StringBuilderOps` on an existing `StringBuilder`, so if someone wants to access it from the outside, they can use that. (That possibility is new and wasn't present in Gizmo 1.x.) So I can just remove `instance()` if you agree.

Ladicek: Done, removed the method.

src/main/java/io/quarkus/gizmo2/creator/ops/StringBuilderOps.java
Ladicek: Done.

src/main/java/io/quarkus/gizmo2/creator/ops/StringBuilderOps.java
Ladicek: Removed the method in the end.

mkouba (approved):

dmlloyd (approved):

PR #278, opened 2025-03-31
add utility to generate equals, hashCode and toString

### LINKED ISSUE #258: Gizmo2: port EqualsHashCodeToStringGenerator
See https://github.com/quarkusio/gizmo/blob/main/src/main/java/io/quarkus/gizmo/Gizmo.java#L1192

Ladicek:
This is an equivalent of `EqualsHashCodeToStringGenerator` from Gizmo 1.x.

Resolves #258

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -41,16 +41,33 @@
         <maven.compiler.release>17</maven.compiler.release>
 
         <version.classfile>24.0</version.classfile>
+        <version.smallrye-common>2.11.0</version.smallrye-common>
         <version.junit>5.12.1</version.junit>
 
         <version.nexus-staging-plugin>1.7.0</version.nexus-staging-plugin>
     </properties>
+
+    <dependencyManagement>
+        <dependencies>
+            <dependency>
+                <groupId>io.smallrye.common</groupId>
+                <artifactId>smallrye-common-bom</artifactId>
+                <version>${version.smallrye-common}</version>
+                <type>pom</type>
+                <scope>import</scope>
+            </dependency>
+        </dependencies>
+    </dependencyManagement>
 
     <dependencies>
         <dependency>
             <groupId>io.github.dmlloyd</groupId>
             <artifactId>jdk-classfile-backport</artifactId>
             <version>${version.classfile}</version>
+        </dependency>
+        <dependency>
+            <groupId>io.smallrye.common</groupId>
+            <artifactId>smallrye-common-constraint</artifactId>
         </dependency>
         <dependency>
             <groupId>org.junit.jupiter</groupId>
--- a/src/main/java/io/quarkus/gizmo2/creator/BlockCreator.java
+++ b/src/main/java/io/quarkus/gizmo2/creator/BlockCreator.java
@@ -2114,6 +2114,15 @@ public sealed interface BlockCreator extends SimpleTyped permits BlockCreatorImp
     void ifInstanceOf(Expr obj, ClassDesc type, BiConsumer<BlockCreator, Expr> ifTrue);
 
     /**
+     * If the given object is <em>not</em> an instance of the given type, then execute the given block.
+     *
+     * @param obj the object to test (must not be {@code null})
+     * @param type the type to check for (must not be {@code null})
+     * @param ifFalse the builder for a block to run if the type did not match (must not be {@code null})
+     */
+    void ifNotInstanceOf(Expr obj, ClassDesc type, Consumer<BlockCreator> ifFalse);
+
+    /**
      * If the given object is an instance of the given type, then execute the first block with the narrowed object,
      * otherwise execute the other block.
      *
--- a/src/main/java/io/quarkus/gizmo2/creator/BlockCreator.java
+++ b/src/main/java/io/quarkus/gizmo2/creator/BlockCreator.java
@@ -2509,52 +2509,64 @@ public sealed interface BlockCreator extends SimpleTyped permits BlockCreatorImp
     // useful helpers/utilities
 
     /**
-     * {@return the hash code of the given expression}
+     * Generates call to one of the static methods to compute a hash code of given expression.
+     * That is:
+     * <ul>
+     * <li>for reference types, {@code Objects.hashCode(expr)}</li>
+     * <li>for primitive types, the static {@code hashCode(expr)} method on the corresponding wrapper class</li>
+     * </ul>
+     *
      * @param expr the expression, which can be of any type (must not be {@code null})
+     * @return an {@code int} expression representing the hash code of given expression (not {@code null})
      */
     Expr exprHashCode(Expr expr);
 
     /**
-     * Determine the equality of the given objects using <em>value equality</em>
-     * rather than <em>reference equality</em>.
-     * This is analogous to using {@link Object#equals(Object)} to compare references.
-     * For non-reference types, this would be equivalent to {@link #eq(Expr, Expr)}.
+     * Generates call to the {@code Objects#equals(a, b)} method if at least one
+     * of the given expressions is of a reference type (boxing the other if primitive).
+     * If both expressions are of a primitive type, this is equivalent to {@link #eq(Expr, Expr)}.
      *
      * @param a the first expression (must not be {@code null})
      * @param b the second expression (must not be {@code null})
-     * @return a {@code boolean} expression representing the equality (or lack thereof) between the two values
+     * @return a {@code boolean} expression representing the equality between the two values (not {@code null})
      */
     Expr exprEquals(Expr a, Expr b);
 
     /**
-     * Determine the equality of the given objects using <em>value equality</em>
-     * rather than <em>reference equality</em>.
-     * This is analogous to using {@link Object#equals(Object)} to compare references.
-     * For non-reference types, this would be equivalent to {@link #eq(Expr, Expr)}.
+     * Generates call to one of the {@code String#valueOf(expr)} overloads, based on the type of the argument.
      *
-     * @param a the first expression (must not be {@code null})
-     * @param b the second expression (must not be {@code null})
-     * @return a {@code boolean} expression representing the equality (or lack thereof) between the two values
-     */
-    default Expr exprEquals(Expr a, ConstantDesc b) {
-        return exprEquals(a, Constant.of(b));
-    }
-
-    /**
-     * {@return a string expression for the given expression}
      * @param expr the expression, which can be of any type
+     * @return a {@code String} expression representing the string value of given expression (not {@code null})
      */
     Expr exprToString(Expr expr);
 
     /**
-     * Generate a call to one of the {@code Arrays#equals(a, b)} overloads, based on the type of the first argument.
-     * The argument types must be the same.
+     * Generates call to one of the {@code Arrays#hashCode(expr)} overloads, based on the type of the argument,
+     * or to {@code Arrays#deepHashCode(expr)} in case of multidimensional arrays.
+     *
+     * @param expr the array instance (must not be {@code null})
+     * @return an {@code int} expression representing the hash code of given array (not {@code null})
+     */
+    Expr arrayHashCode(Expr expr);
+
+    /**
+     * Generates call to one of the {@code Arrays#equals(a, b)} overloads, based on the type of the first argument,
+     * or to {@code Arrays#deepEquals(a, b)} in case of multidimensional arrays.
      *
      * @param a the first array instance (must not be {@code null})
      * @param b the second array instance (must not be {@code null})
-     * @return the expression representing the result of the equality check (not {@code null})
+     * @return a {@code boolean} expression representing the equality between the two arrays (not {@code null})
      */
     Expr arrayEquals(Expr a, Expr b);
+
+    /**
+     * Generates call to one of the {@code Arrays#toString(expr)} overloads, based on the type of the argument,
+     * or to {@code Arrays#deepToString(expr)} in case of multidimensional arrays.
+     *
+     * @param expr the array instance (must not be {@code null})
+     * @return a {@code String} expression representing the string value of given array (not {@code null})
+     */
+    Expr arrayToString(Expr expr);
 
     /**
      * {@return a convenience wrapper for accessing instance methods of {@link Object}}
@@ -2584,6 +2596,7 @@ public sealed interface BlockCreator extends SimpleTyped permits BlockCreatorImp
      * {@return a convenience wrapper for accessing instance methods of {@link Collection}}
      * @param receiver the instance to invoke upon (must not be {@code null})
      */
+
     default CollectionOps withCollection(Expr receiver) {
         return new CollectionOps(this, receiver);
     }
--- a/src/main/java/io/quarkus/gizmo2/creator/BlockCreator.java
+++ b/src/main/java/io/quarkus/gizmo2/creator/BlockCreator.java
@@ -2138,7 +2138,7 @@ public sealed interface BlockCreator extends SimpleTyped permits BlockCreatorImp
      * @param cond the boolean condition expression (must not be {@code null})
      * @param whenFalse the builder for a block to execute if the condition is false (must not be {@code null})
      */
-    void unless(Expr cond, Consumer<BlockCreator> whenFalse);
+    void ifNot(Expr cond, Consumer<BlockCreator> whenFalse);
 
     /**
      * A general {@code if}-{@code else} conditional.
--- a/src/main/java/io/quarkus/gizmo2/creator/ClassCreator.java
+++ b/src/main/java/io/quarkus/gizmo2/creator/ClassCreator.java
@@ -2,6 +2,7 @@ package io.quarkus.gizmo2.creator;
 
 import java.lang.constant.ClassDesc;
 import java.lang.constant.MethodTypeDesc;
+import java.util.List;
 import java.util.function.Consumer;
 
 import io.github.dmlloyd.classfile.Signature;
@@ -11,6 +12,7 @@ import io.quarkus.gizmo2.desc.ConstructorDesc;
 import io.quarkus.gizmo2.desc.FieldDesc;
 import io.quarkus.gizmo2.desc.MethodDesc;
 import io.quarkus.gizmo2.impl.ClassCreatorImpl;
+import io.quarkus.gizmo2.impl.EqualsHashCodeToStringGenerator;
 import io.quarkus.gizmo2.impl.Util;
 
 /**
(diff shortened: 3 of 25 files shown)


Ladicek:
Commits are supposed to be reviewed in isolation.

src/main/java/io/quarkus/gizmo2/impl/EqualsHashCodeToStringGenerator.java
```diff
@@ -0,0 +1,297 @@
...
+                });
+
+                Expr otherCast = b0.define("other", b0.cast(other, thisClass));
+                for (FieldDesc field : fields) {
+                    if (!field.owner().equals(thisClass)) {
+                        throw new IllegalArgumentException("Field does not belong to " + thisClass.displayName() + ": " + field);
+                    }
+
+                    Expr thisValue = b0.define("thisValue", b0.get(mc.this_().field(field)));
+                    Expr thatValue = b0.define("thatValue", b0.get(otherCast.field(field)));
```
dmlloyd, line 63: I would type these as `LocalVar`.

src/main/java/io/quarkus/gizmo2/impl/EqualsHashCodeToStringGenerator.java
```diff
@@ -0,0 +1,297 @@
...
+                    switch (field.type().descriptorString()) {
```
dmlloyd, line 64: These repeated cases where it's a `switch` of a descriptor string, but then also an `if` tree, might be better written (sometimes at least) as `switch (descriptorString.charAt(0)`, with a sub-switch for the `[` case. This would eliminate the extra `if` tree and might end up looking marginally cleaner, maybe?

src/main/java/io/quarkus/gizmo2/impl/EqualsHashCodeToStringGenerator.java
```diff
@@ -0,0 +1,297 @@
...
+                            // this is consistent with Arrays.equals() and it's also what IntelliJ generates
+                            Expr thisBits = b0.invokeStatic(doubleToLongBits, thisValue);
+                            Expr thatBits = b0.invokeStatic(doubleToLongBits, thatValue);
+                            b0.if_(b0.ne(thisBits, thatBits), b1 -> {
+                                b1.return_(false);
+                            });
+                        }
+                        // boolean[]
+                        case "[Z" ->
+                            b0.unless(b0.invokeStatic(booleanArrayEquals, thisValue, thatValue), b1 -> {
```
dmlloyd, line 90: Since the method descriptors are already being acquired within the above method, it might be simpler overall to combine similar ccases and have it be like: `b0.invokeStatic(ClassMethodDesc.of(CD_Arrays, "equals", field.type(), field.type())` without the declarations above. WDYT?

src/main/java/io/quarkus/gizmo2/impl/Neg.java
```diff
@@ -26,12 +26,12 @@ public boolean bound() {
...
 
     public void writeCode(final CodeBuilder cb, final BlockCreatorImpl block) {
-        switch (typeKind()) {
+        switch (typeKind().asLoadable()) {
             case INT -> cb.ineg();
             case LONG -> cb.lneg();
             case FLOAT -> cb.fneg();
             case DOUBLE -> cb.dneg();
-            default -> throw new IllegalStateException();
+            default -> throw new IllegalStateException("Cannot negate non-primitive expression: " + a);
```
dmlloyd, line 34: For validations, we should do them up front (in `BlockCreatorImpl#neg()` in this case). Otherwise what happens is that the exception doesn't come up until much later and we lose the line number of the problem.

src/main/java/io/quarkus/gizmo2/creator/BlockCreator.java
```diff
@@ -2113,6 +2113,15 @@ default Expr invokeDynamic(DynamicCallSiteDesc callSiteDesc, Expr... args) {
...
     void ifInstanceOf(Expr obj, ClassDesc type, BiConsumer<BlockCreator, Expr> ifTrue);
 
+    /**
+     * If the given object is <em>not</em> an instance of the given type, then execute the given block.
+     *
+     * @param obj the object to test (must not be {@code null})
+     * @param type the type to check for (must not be {@code null})
+     * @param ifFalse the builder for a block to run if the type did not match (must not be {@code null})
+     */
+    void ifNotInstanceOf(Expr obj, ClassDesc type, Consumer<BlockCreator> ifFalse);
```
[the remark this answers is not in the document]
Ladicek: I would actually much rather rename `unless()` to `ifNot()`. I originally had `bc.if_(bc.logicalNot(...))` and only after looking at how alien the generated code looked after decompilation I went looking for alternatives and found `unless()` :-)

Didn't want to do it in this PR though.

src/main/java/io/quarkus/gizmo2/impl/EqualsHashCodeToStringGenerator.java
Ladicek: Makes sense, will do.

src/main/java/io/quarkus/gizmo2/impl/EqualsHashCodeToStringGenerator.java
Ladicek: That didn't occur to me at all, makes sense. Will do.

src/main/java/io/quarkus/gizmo2/impl/EqualsHashCodeToStringGenerator.java
```diff
@@ -0,0 +1,297 @@
...
+        ClassDesc thisClass = cc.type();
+
+        cc.method("equals", mc -> {
+            mc.public_();
+            mc.returning(boolean.class);
+            ParamVar other = mc.parameter("other", Object.class);
+            mc.body(b0 -> {
+                b0.if_(b0.eq(mc.this_(), other), b1 -> {
+                    b1.return_(true);
+                });
```
mkouba, line 51:
```suggestion
                b0.if_(b0.eq(mc.this_(), other), BlockCreator::returnTrue);
```
Maybe?

src/main/java/io/quarkus/gizmo2/impl/Neg.java
Ladicek: I didn't check if this is validate before, I just stumbled upon an ISE without a message and figured I'll write something at least. I'll check how that's done. If there's an upfront validation, I guess we could throw `AssertionError` here or something like that.

src/main/java/io/quarkus/gizmo2/impl/EqualsHashCodeToStringGenerator.java
Ladicek: Sounds reasonable, will do.

src/main/java/io/quarkus/gizmo2/creator/BlockCreator.java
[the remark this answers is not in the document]
dmlloyd: FWIW `unless` comes from Perl. I'm not sure if that's good or bad though. ;-)

src/main/java/io/quarkus/gizmo2/impl/Neg.java
dmlloyd: ISE is commonly used for invalid switch states IIRC. There's also an `impossibleSwitchCase()` in `smallrye-common-constraint` that might be a good choice here.

Ladicek: That makes sense.

src/main/java/io/quarkus/gizmo2/creator/BlockCreator.java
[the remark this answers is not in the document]
dmlloyd: Maybe we want cross-referencing `@see` tags in `if` and `unless`?

src/main/java/io/quarkus/gizmo2/creator/ClassCreator.java
```diff
@@ -259,4 +261,83 @@ default ConstructorDesc defaultConstructor() {
...
+     * </li>
+     * <li>A closing parenthesis {@code ')'} is appended.</li>
+     * <li>The {@code StringBuilder.toString()} outcome is returned.</li>
+     * </ol>
+     * <p>
+     * If one of the fields doesn't belong to this class, an exception is thrown.
+     *
+     * @param fields fields to consider in the {@code toString} methods (must not be {@code null})
+     */
+    default void generateNaiveToString(List<FieldDesc> fields) {
```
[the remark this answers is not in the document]
mkouba: Yes please!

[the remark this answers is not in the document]
Ladicek: Will do.

src/main/java/io/quarkus/gizmo2/impl/EqualsHashCodeToStringGenerator.java
Ladicek: Done.

src/main/java/io/quarkus/gizmo2/impl/Neg.java
Ladicek: Done. Added argument validation to the constructor, similarly to `Rel`.

src/main/java/io/quarkus/gizmo2/impl/EqualsHashCodeToStringGenerator.java
Ladicek: Done.

src/main/java/io/quarkus/gizmo2/creator/ClassCreator.java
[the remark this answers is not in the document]
Ladicek: Done.

src/main/java/io/quarkus/gizmo2/impl/BlockCreatorImpl.java
```diff
@@ -1073,17 +1078,18 @@ public void yield(final Expr val) {
...
+            case 'Z' -> invokeStatic(MethodDesc.of(Boolean.class, "hashCode", int.class, boolean.class), expr);
+            case 'B' -> invokeStatic(MethodDesc.of(Byte.class, "hashCode", int.class, byte.class), expr);
+            case 'S' -> invokeStatic(MethodDesc.of(Short.class, "hashCode", int.class, short.class), expr);
+            case 'I' -> invokeStatic(MethodDesc.of(Integer.class, "hashCode", int.class, int.class), expr);
+            case 'J' -> invokeStatic(MethodDesc.of(Long.class, "hashCode", int.class, long.class), expr);
+            case 'F' -> invokeStatic(MethodDesc.of(Float.class, "hashCode", int.class, float.class), expr);
+            case 'D' -> invokeStatic(MethodDesc.of(Double.class, "hashCode", int.class, double.class), expr);
+            case 'C' -> invokeStatic(MethodDesc.of(Character.class, "hashCode", int.class, char.class), expr);
+            case 'L', '[' -> invokeStatic(MethodDesc.of(Objects.class, "hashCode", int.class, Object.class), expr);
+            default -> throw impossibleSwitchCase(expr);
```
dmlloyd, line 1092: Did you discover some problem with using type kinds that prompted this change?

src/main/java/io/quarkus/gizmo2/impl/BlockCreatorImpl.java
```diff
@@ -1108,20 +1114,51 @@ public Expr exprToString(final Expr expr) {
...
+        }
+    }
+
     public Expr arrayEquals(final Expr a, final Expr b) {
-        ClassDesc type = switch (TypeKind.from(a.type().componentType())) {
-            case REFERENCE -> CD_Object.arrayType();
-            default -> a.type();
-        };
-        return invokeStatic(MethodDesc.of(Arrays.class, "equals", MethodTypeDesc.of(CD_boolean, type, type)), a, b);
+        if (!a.type().isArray() || !b.type().isArray() || !a.type().equals(b.type())) {
```
dmlloyd, line 1137: This isn't strictly true; I could compare a `String[]` and an `Object[]` for equality, right? Maybe it should just check for equality of the `TypeKind` instead.

Also, we have some validation methods in `Item` which are normally invoked from the constructor of `Item` subclasses, but I wonder if they should be centralized (maybe even using JBoss Logging message formatting stuff?).

src/main/java/io/quarkus/gizmo2/impl/EqualsHashCodeToStringGenerator.java
```diff
@@ -0,0 +1,146 @@
...
+            mc.returning(int.class);
+            mc.body(b0 -> {
+                if (fields.isEmpty()) {
+                    b0.return_(0);
+                    return;
+                }
+
+                LocalVar result = b0.define("result", Constant.of(1));
+                for (FieldDesc field : fields) {
+                    if (!field.owner().equals(thisClass)) {
```
dmlloyd, line 99: You could conceivably include fields of a superclass, right?

src/main/java/io/quarkus/gizmo2/impl/Neg.java
```diff
@@ -4,13 +4,19 @@
...
 import io.quarkus.gizmo2.Expr;
 
+import static io.smallrye.common.constraint.Assert.impossibleSwitchCase;
+
 final class Neg extends Item {
     private final Item a;
 
     Neg(final Expr a) {
         this.a = (Item) a;
+        if (a.typeKind() == TypeKind.REFERENCE || a.typeKind() == TypeKind.VOID) {
```
dmlloyd, line 17: Maybe add `BOOLEAN` in there too?

src/main/java/io/quarkus/gizmo2/impl/Neg.java
```diff
@@ -26,12 +32,12 @@ public boolean bound() {
...
 
     public void writeCode(final CodeBuilder cb, final BlockCreatorImpl block) {
-        switch (typeKind()) {
+        switch (typeKind().asLoadable()) {
             case INT -> cb.ineg();
             case LONG -> cb.lneg();
             case FLOAT -> cb.fneg();
             case DOUBLE -> cb.dneg();
-            default -> throw new IllegalStateException();
+            default -> throw impossibleSwitchCase(a);
```
dmlloyd, line 40: Normally you'd put in the case that was impossible here, i.e. `typeKind().asLoadable()`.

src/main/java/io/quarkus/gizmo2/impl/Rel.java
```diff
@@ -59,7 +60,7 @@ public void writeCode(final CodeBuilder cb, final BlockCreatorImpl block) {
         switch (typeKind().asLoadable()) {
             case INT -> kind.if_icmp.accept(cb, true_);
             case REFERENCE -> kind.if_acmp.accept(cb, true_);
-            default -> throw new IllegalStateException();
+            default -> throw impossibleSwitchCase(a + " " + kind + " " + b);
```
dmlloyd, line 63: Same here (i.e. `typeKind().asLoadable()`).

src/main/java/io/quarkus/gizmo2/creator/BlockCreator.java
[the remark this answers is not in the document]
dmlloyd: Well, either this should be `unlessInstanceOf`, or else `unless` should change to `ifNot`. We shouldn't it have two ways to spell it.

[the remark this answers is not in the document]
mkouba: > Well, either this should be `unlessInstanceOf`, or else `unless` should change to `ifNot`. We shouldn't it have two ways to spell it.

+100. Now, how do we decide which one to choose?

`ifNot`/`ifNotInstanceOf`
1. easier to find when using autocompletion (because you start with `if`)
2. you can save 1 char ;-)
3. Java does have `if`

`unless`/`unless`/`ifNotInstanceOf`
1. single word, one less capital letter
2. Java does not have `unless`

BTW in Gizmo1 we have (among others) `ifTrue()` and `ifFalse()`.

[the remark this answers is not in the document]
dmlloyd: Ah, good, a third option. 😉

src/main/java/io/quarkus/gizmo2/impl/BlockCreatorImpl.java
Ladicek: Not really. I originally had code to deal with multidimensionals array here too, so I had to differentiate between objects and arrays. I'll put the code back to its original form.

src/main/java/io/quarkus/gizmo2/impl/Neg.java
Ladicek: `BOOLEAN` is an `INT` and I've accidentally verified before that `logicalNot(booleanExpr)` works (even if it generates strange bytecode), so I don't think we strictly _need_ it. It's probably a good idea though.

src/main/java/io/quarkus/gizmo2/impl/EqualsHashCodeToStringGenerator.java
Ladicek: I didn't think about that, but you're right. I'm not really keen on removing the check, but I don't see a better way...

src/main/java/io/quarkus/gizmo2/impl/Neg.java
Ladicek: Since `typeKind()` boils down to `a.typeKind()`, I figured reporting the whole `a` would be more informative here.

src/main/java/io/quarkus/gizmo2/creator/BlockCreator.java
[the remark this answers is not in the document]
mkouba: Ok, let's choose one option for now and we could change our mind afterwards.

I vote for `ifNot/ifNotInstanceOf` (although it was a tough decision, `ifNot`/`ifNotInstanceOf` scores 51/100 and `unless`/`unlessInstanceOf` scores 49/100 ;-).

src/main/java/io/quarkus/gizmo2/impl/Neg.java
Ladicek: Sorry, my bad, `logicalNot()` doesn't translate to `Neg`. Will add a `BOOLEAN` case.

src/main/java/io/quarkus/gizmo2/creator/BlockCreator.java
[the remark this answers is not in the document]
Ladicek: I'm like 2/3 for `ifNot` and 1/3 for `unless`, if you ask me :-)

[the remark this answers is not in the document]
mkouba: So _unless_ @dmlloyd has some serious concerns about `ifNot` _then_ pls rename those members...

src/main/java/io/quarkus/gizmo2/impl/EqualsHashCodeToStringGenerator.java
Ladicek: So I'm thinking handling superclasses is out of scope, at least for now. The best way to handle superclasses would be to call `equals()` (or `hashCode()`; not sure about `toString()`) on the superclass first, and then handle the fields of this class.

src/main/java/io/quarkus/gizmo2/creator/BlockCreator.java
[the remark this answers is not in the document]
Ladicek: Renamed `unless` to `ifNot` :-)

src/main/java/io/quarkus/gizmo2/impl/BlockCreatorImpl.java
Ladicek: Done.

src/main/java/io/quarkus/gizmo2/impl/BlockCreatorImpl.java
Ladicek: Done. Moved the validation methods from `Item` to a new class `Preconditions` (that's the name I use in SmallRye Fault Tolerance, can use a different one if you prefer). Didn't use the JBoss Logging formatting stuff yet, maybe later?

src/main/java/io/quarkus/gizmo2/impl/Neg.java
Ladicek: Done.

src/main/java/io/quarkus/gizmo2/impl/Neg.java
Ladicek: Changed it to `typeKind().asLoadable()` in the end. Done.

src/main/java/io/quarkus/gizmo2/impl/Rel.java
Ladicek: Done.

src/main/java/io/quarkus/gizmo2/creator/BlockCreator.java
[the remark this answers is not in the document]
dmlloyd: Great!

dmlloyd (approved):

pom.xml
```diff
@@ -52,6 +53,11 @@
             <artifactId>jdk-classfile-backport</artifactId>
             <version>${version.classfile}</version>
         </dependency>
+        <dependency>
```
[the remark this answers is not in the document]
Ladicek: I didn't even know SmallRye Common has a BOM :-) Will change.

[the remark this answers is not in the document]
Ladicek: Done.

PR #281, opened 2025-04-01
improve the `Expr` type hierarchy

Ladicek:
The previous `Expr` type hierarchy looked like this:

```
Expr
├─ Constant
├─ LValueExpr
│  ├─ FieldVar
│  │  ├─ InstanceFieldVar
│  │  └─ StaticFieldVar
│  ├─ LocalVar
│  ├─ ParamVar
│  └─ LValueExprImpl
├─ Var
│  ├─ FieldVar
│  │  ├─ StaticFieldVar
│  │  └─ InstanceFieldVar
│  ├─ LocalVar
│  ├─ ParamVar
│  └─ ThisExpr
└─ Item
   ├─ ThisExpr
   └─ other subclasses
```

The new `Expr` hierarchy looks like this:

```
Expr
├─ Constant
├─ This
│  └─ ThisExpr
├─ Assignable (previously `LValueExpr`)
│  ├─ Var
│  │  ├─ FieldVar
│  │  │  ├─ InstanceFieldVar
│  │  │  └─ StaticFieldVar
│  │  ├─ LocalVar
│  │  └─ ParamVar
│  └─ AssignableImpl (previously `LValueExprImpl`)
└─ Item (many subclasses also implement the interfaces above)
```

All `Var`s are `Assignable`s now. `ThisExpr` is no longer a `Var`,
it has a separate user-facing interface `This`. It no longer has
a `name()` method, which was the only method prescribed by the `Var`
interface. I don't think this is a big loss.


Ladicek:
Marking this as a draft because this is meant mostly for discussion. This is what I mentioned on our call yesterday.

Ladicek:
One more thing: I know the term lvalue is fairly common in the C lingo. I have only found 1 occurence of this term in the JLS (referring to C again). In Rust, the concept of "lvalue expression" is called "place expression", which I find is much more descriptive. Hence that rename.

mkouba:
> Values are Constants and Items.

What exactly does an `Item` represent? Could we find a better name?

Ladicek:
> What exactly does an `Item` represent? Could we find a better name?

That is a _great_ question that I'm not capable of answering :-)

dmlloyd:
An `Item` is an... item... in the linked list of a block. Every expression, constant, variable etc. must extend this common base class to implement various code emitting behaviors. Since it's an implementation class, its name reflects its implementation concept. Items can emit code, and also know how to insert or remove themselves from a linked list.

As for terminology - I guess if `lvalue` isn't appropriate then `place` probably isn't either. I'm looking a the JLS for the most idiomatic word to use for this, and it's a bit tricky honestly. :-) I wonder if we might consider `Assignable` perhaps, or maybe `WritableValue`? One thing I noticed is that the JLS considers an array dereference expression like `foo[123][456]` to be a "variable", which complicates things further.

If you do go with something like `Value` then the issue is that a variable is usable as a "value" by any sane definition of the word; thus `Value` is more synonymous with `Expr` in my opinion. If you want to introduce a subtype whose primary characteristic is that it cannot be assigned to, you'd probably end up with `Unassignable` or `ReadOnlyValue` or something like that.

`this` is not as simple as it appears. In Java it's a special keyword, but in the JVM it's a variable at slot 0. It implements `Var` but not `LValue` just because assigning to `this` is a pretty clear footgun (though implementation-wise it certainly could implement, say, `LocalVar`, and we could just do this without any real negative implementation consequence). But, it *should* be considered a variable because it has a slot. Also, right now the way local variables are allocated is fairly naive in that a variable always lives to the end of its scope (it's somewhat better than Gizmo 1, where this is also technically true, because in Gizmo 1 there is really only one scope in most cases and every value is a variable, but it's still far from optimal). But an idea I was looking at would be to allow the user to constrain the lifetime of some kinds of variable. I might do this by having `Var` extend `AutoCloseable`, or by creating a sibling of `Var` to represent a constrained-lifetime variable like `Scoped` which would be implemented by `this`, local variables, and parameters. Then the user could do e.g.:

```java
    try (LocalVar temp = b0.define("temp", b0.add(foo, 10))) {
        // operate on `temp`
    }
```

After the scope completes, the slot can be reused by nested blocks, subsequent variable definitions, etc. This would also include parameters, and indeed `this`, though in these cases the user would need to explicitly call `close()` on the variable to indicate that it is no longer in scope. A side effect of this idea would be that any "don't care" parameters (including `this`), which are identified as parameters that are not declared before the body of a method, would not get local variable table entries or slots, and those slots would be usable immediately by other local variables.

I also noticed a bit of JavaDoc rework. On "parameters" and "arguments": a thing is a "parameter" if it is part of the formal specification of a method, so this would apply to parameter types in method type specifications, and parameter variables within a method definition. An argument would be the value given for a parameter at a use site, so this would apply to method and constructor calls mainly. I noticed there was one place where you changed "argument" to "parameter" but the correct term would be "argument" in this case.

src/main/java/io/quarkus/gizmo2/creator/ops/ObjectOps.java
```diff
@@ -67,61 +67,61 @@ protected Expr receiver() {
...
+     * Perform an instance invocation of a method with 3 parameters.
      *
      * @param returnType the return type (must not be {@code null})
      * @param name the method name (must not be {@code null})
-     * @param arg0Type the first argument type (must not be {@code null})
-     * @param arg1Type the second argument type (must not be {@code null})
-     * @param arg2Type the third argument type (must not be {@code null})
+     * @param param0Type the first parameter type (must not be {@code null})
+     * @param param1Type the second parameter type (must not be {@code null})
+     * @param param2Type the third parameter type (must not be {@code null})
```
dmlloyd, line 76: I guess you're correct here; these are indeed the parameter types and not argument types.

Ladicek:
> As for terminology - I guess if `lvalue` isn't appropriate then `place` probably isn't either. I'm looking a the JLS for the most idiomatic word to use for this, and it's a bit tricky honestly. :-) I wonder if we might consider `Assignable` perhaps, or maybe `WritableValue`? One thing I noticed is that the JLS considers an array dereference expression like `foo[123][456]` to be a "variable", which complicates things further.

Indeed the JLS doesn't have the best terminology for this. I went with `Value` and `Place` because those seem most descriptive to me. `Assignable` instead of `Place` makes sense.

> If you do go with something like `Value` then the issue is that a variable is usable as a "value" by any sane definition of the word; thus `Value` is more synonymous with `Expr` in my opinion. If you want to introduce a subtype whose primary characteristic is that it cannot be assigned to, you'd probably end up with `Unassignable` or `ReadOnlyValue` or something like that.

It is true that places are also always values. I guess I could get rid of the `Value` interface and let all values just extend `Expr` like before. Not a huge deal.

> `this` is not as simple as it appears. In Java it's a special keyword, but in the JVM it's a variable at slot 0. It implements `Var` but not `LValue` just because assigning to `this` is a pretty clear footgun (though implementation-wise it certainly could implement, say, `LocalVar`, and we could just do this without any real negative implementation consequence). But, it _should_ be considered a variable because it has a slot. Also, right now the way local variables are allocated is fairly naive in that a variable always lives to the end of its scope (it's somewhat better than Gizmo 1, where this is also technically true, because in Gizmo 1 there is really only one scope in most cases and every value is a variable, but it's still far from optimal). But an idea I was looking at would be to allow the user to constrain the lifetime of some kinds of variable. I might do this by having `Var` extend `AutoCloseable`, or by creating a sibling of `Var` to represent a constrained-lifetime variable like `Scoped` which would be implemented by `this`, local variables, and parameters. Then the user could do e.g.:
> 
> ```java
>     try (LocalVar temp = b0.define("temp", b0.add(foo, 10))) {
>         // operate on `temp`
>     }
> ```
> 
> After the scope completes, the slot can be reused by nested blocks, subsequent variable definitions, etc. This would also include parameters, and indeed `this`, though in these cases the user would need to explicitly call `close()` on the variable to indicate that it is no longer in scope. A side effect of this idea would be that any "don't care" parameters (including `this`), which are identified as parameters that are not declared before the body of a method, would not get local variable table entries or slots, and those slots would be usable immediately by other local variables.

So I guess I wouldn't really want to allow _anything_ you describe here to `this`, even if it makes sense from the JVMS perspective, because it doesn't on the JLS level. In fact, I guess I wouldn't even allow closing variable scopes manually -- a variable is scoped to the block in which it was declared, period.

If we wanted to shrink the number of local variables, I think we should implement a limited form of register allocation (we wouldn't need spilling for example). This was more complex in Gizmo 1.x, where the IR was a tree of instructions, but maybe in Gizmo 2.x, we have a linear IR? I don't think so, but I have ignored all about that for now, so I might easily be wrong.

mkouba:
> As for terminology - I guess if `lvalue` isn't appropriate then `place` probably isn't either. I'm looking a the JLS for the most idiomatic word to use for this, and it's a bit tricky honestly. :-) I wonder if we might consider `Assignable` perhaps, or maybe `WritableValue`?

`Assignable` seems to be quite intuitive in this case...

dmlloyd:
> > If you do go with something like `Value` then the issue is that a variable is usable as a "value" by any sane definition of the word; thus `Value` is more synonymous with `Expr` in my opinion. If you want to introduce a subtype whose primary characteristic is that it cannot be assigned to, you'd probably end up with `Unassignable` or `ReadOnlyValue` or something like that.
> 
> It is true that places are also always values. I guess I could get rid of the `Value` interface and let all values just extend `Expr` like before. Not a huge deal.

I think it's OK to have the subtype for non-writable values, but it should just be clear what it is. I'm also open to the idea of keeping it as-is, or renaming `Expr` to `Value`, or any combination. The hierarchy underwent several changes already since its inception.

> So I guess I wouldn't really want to allow _anything_ you describe here to `this`, even if it makes sense from the JVMS perspective, because it doesn't on the JLS level. In fact, I guess I wouldn't even allow closing variable scopes manually -- a variable is scoped to the block in which it was declared, period.
> 
> If we wanted to shrink the number of local variables, I think we should implement a limited form of register allocation (we wouldn't need spilling for example). This was more complex in Gizmo 1.x, where the IR was a tree of instructions, but maybe in Gizmo 2.x, we have a linear IR? I don't think so, but I have ignored all about that for now, so I might easily be wrong.

Well, remember, this isn't an optimizing compiler, this is a bytecode writing library. While it is useful to reference the JLS to make sure things are reasonably intuitive, and I don't object to optimizing some low-hanging fruit, I do think the user can have *some* responsibility for the code they write at this level. Tracking variable lifetimes automatically may be pretty complex due to blocks which split and join and other things. Letting the user *optionally* be explicit about it would keep things much simpler on the implementation side in that regard.

The key is to find the right balance here. Right now, the user would have to create nested blocks to establish a nested variable scope, and that is a bit heavier than `try`-with-resources, which is what prompted the idea. Also it could be occasionally useful to explicitly interleave variable lifetimes.

dmlloyd:
> > An `Item` is an... item... in the linked list of a block. Every expression, constant, variable etc. must extend this common base class to implement various code emitting behaviors.
> 
> Hm, so maybe something like `CodeItem`?

Maybe. There was a time when `Item` didn't implement `Expr`, to support things that weren't expressions (like `goto`, `return`, `if`, etc.). Instead I had a subclass of `Item` called `ExprImpl` which implemented the necessary behaviors. But then I changed these things to be `Expr` with type `void`, which unified a lot of things. So `Item` could feasibly be renamed to `ExprImpl`, I guess, though it doesn't quite fit the characteristics of an `Item`.

`CodeItem` is also not a perfect fit, because an `Item` does not always translate into some piece of code. The characteristics of an `Item` are:

* Can generate code, if applicable
* Can specify a particular insertion or removal behavior (wrt the linked list)
* Provides infrastructure for dependency-based list traversal
* Provides verification and validation infrastructure

But all of this stuff is non-user-facing so I think it is not super important to nail the naming of it any particular way, as long as the documentation is adequate (n.b.: it is not currently adequate 😅).

Ladicek:
> > So I guess I wouldn't really want to allow _anything_ you describe here to `this`, even if it makes sense from the JVMS perspective, because it doesn't on the JLS level. In fact, I guess I wouldn't even allow closing variable scopes manually -- a variable is scoped to the block in which it was declared, period.
> > If we wanted to shrink the number of local variables, I think we should implement a limited form of register allocation (we wouldn't need spilling for example). This was more complex in Gizmo 1.x, where the IR was a tree of instructions, but maybe in Gizmo 2.x, we have a linear IR? I don't think so, but I have ignored all about that for now, so I might easily be wrong.
> 
> Well, remember, this isn't an optimizing compiler, this is a bytecode writing library. While it is useful to reference the JLS to make sure things are reasonably intuitive, and I don't object to optimizing some low-hanging fruit, I do think the user can have _some_ responsibility for the code they write at this level. Tracking variable lifetimes automatically may be pretty complex due to blocks which split and join and other things. Letting the user _optionally_ be explicit about it would keep things much simpler on the implementation side in that regard.
> 
> The key is to find the right balance here. Right now, the user would have to create nested blocks to establish a nested variable scope, and that is a bit heavier than `try`-with-resources, which is what prompted the idea. Also it could be occasionally useful to explicitly interleave variable lifetimes.

Right, so I'd like to point out that I don't believe we need any of those optimizations I was thinking about. I didn't measure it, so I can easily be wrong here, but the only thing this might affect is the size of the class file (and marginally at best). I find it highly unlikely to have any measurable effect on the output of an optimizing compiler like C2 or Graal.

That said, I also believe Gizmo is the right place to make such optimizations, if we ever find them necessary. I wouldn't want Gizmo users to do anything like what we're discussing. And I doubly wouldn't want them to do it to `this`.

Or, in other words: Gizmo is not just a bytecode writing library, it is a bytecode writing library for mere mortals. I find it highly appropriate to restrict access to powerful but dangerous stuff, especially if we cannot detect that anything went wrong (which I don't think we would be able to do in this particular case).

Ladicek:
Renamed `Place` to `Assignable` and removed `Value` -- all values directly extend `Expr` now.

src/main/java/io/quarkus/gizmo2/impl/ThisExpr.java
```diff
@@ -3,9 +3,8 @@
 import java.lang.constant.ClassDesc;
 
 import io.github.dmlloyd.classfile.CodeBuilder;
-import io.quarkus.gizmo2.Var;
 
-public final class ThisExpr extends Item implements Var {
+public final class ThisExpr extends Item {
```
dmlloyd, line 7: I still don't like this. With this change, `this` is the only expression that can always be reused many times without having some other kind of involved subtype to indicate that fact. There should be some kind of interface for it. Maybe even just a `This extends Expr`.

Ladicek: That actually makes a lot of sense! I'll fix that.

Ladicek: Done.

dmlloyd:
I guess this one can be closed now?

PR #284, opened 2025-04-03
Fix StaticFieldVarImpl#emitSet(), add access methods to toBlockCreator

mkouba:
- add BlockCreator#readStaticField() and BlockCreator#writeStaticField()
- also add BlockCreator#ifNull() and BlockCreator#ifNotNull()

I'd like to add an `if-null-else`/`if-not-null-else` variants but I don't like the current way we handle the `Else` case (i.e. another method with a `whenFalse` param). Something like this would be more convenient IMO:
```java
bc.ifNull(input, bc1 -> {
   // null case
}).else_(bc2 -> {
   // non-null case
});
```

In Gizmo1 we have a [special construct for `if-then-else`](https://github.com/quarkusio/gizmo/blob/main/USAGE.adoc#if-statements) so maybe we could try to do something similar...

### DIFF
--- a/src/main/java/io/quarkus/gizmo2/creator/BlockCreator.java
+++ b/src/main/java/io/quarkus/gizmo2/creator/BlockCreator.java
@@ -35,6 +35,7 @@ import io.quarkus.gizmo2.creator.ops.SetOps;
 import io.quarkus.gizmo2.creator.ops.StringBuilderOps;
 import io.quarkus.gizmo2.creator.ops.StringOps;
 import io.quarkus.gizmo2.desc.ConstructorDesc;
+import io.quarkus.gizmo2.desc.FieldDesc;
 import io.quarkus.gizmo2.desc.MethodDesc;
 import io.quarkus.gizmo2.impl.BlockCreatorImpl;
 import io.quarkus.gizmo2.impl.Util;
@@ -2159,6 +2160,26 @@ public sealed interface BlockCreator extends SimpleTyped permits BlockCreatorImp
     void ifElse(Expr cond, Consumer<BlockCreator> whenTrue, Consumer<BlockCreator> whenFalse);
 
     /**
+     * An {@code if (obj == null)} conditional.
+     *
+     * @param obj the object reference to test (must not be {@code null})
+     * @param whenTrue the builder for a block to execute if the object reference is null (must not be {@code null})
+     */
+    default void ifNull(Expr obj, Consumer<BlockCreator> whenTrue) {
+        if_(eq(obj, Constant.ofNull(obj.type())), whenTrue);
+    }
+
+    /**
+     * An {@code if (obj != null)} conditional.
+     *
+     * @param obj the object reference to test (must not be {@code null})
+     * @param whenTrue the builder for a block to execute if the object reference is not null (must not be {@code null})
+     */
+    default void ifNotNull(Expr obj, Consumer<BlockCreator> whenTrue) {
+        if_(ne(obj, Constant.ofNull(obj.type())), whenTrue);
+    }
+    
+    /**
      * Construct a {@code switch} statement for {@code enum} constants.
      *
      * @param val the value to switch on (must not be {@code null})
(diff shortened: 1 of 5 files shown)


src/main/java/io/quarkus/gizmo2/creator/BlockCreator.java
```diff
@@ -2158,6 +2159,22 @@ default Expr invokeDynamic(DynamicCallSiteDesc callSiteDesc, Expr... args) {
      */
     void ifElse(Expr cond, Consumer<BlockCreator> whenTrue, Consumer<BlockCreator> whenFalse);
 
+    /**
+     * An {@code if (obj == null)} conditional.
+     *
+     * @param obj the object reference to test (must not be {@code null})
+     * @param whenTrue the builder for a block to execute if the object reference is null (must not be {@code null})
+     */
+    void ifNull(Expr obj, Consumer<BlockCreator> whenTrue);
```
dmlloyd, line 2168: Why is this not a default method that calls `if(eq(obj, Constant.ofNull(obj.type())), whenTrue)`?

src/main/java/io/quarkus/gizmo2/impl/BlockCreatorImpl.java
```diff
@@ -902,6 +902,18 @@ private void doIf(final Expr cond, final Consumer<BlockCreator> whenTrue, final
         }
         doIfInsn(CD_void, cond, wt, wf);
     }
+    
+    private void doIfRef(final Expr ref, final Kind kind, final Consumer<BlockCreator> whenTrue, final Consumer<BlockCreator> whenFalse) {
```
dmlloyd, line 906: None of this should be necessary, the existing `if` code should be able to detect and optimize this case.

Ladicek:
> add BlockCreator#readStaticField() and BlockCreator#writeStaticField()

Given we currently have `BlockCreator.get()` and `set()`, how about we call these `getStatic()` and `setStatic()`? Seems more consistent.

I originally thought it would be enough to add `Expr.staticField()` to `BlockCreator`, but I agree this is better.

src/main/java/io/quarkus/gizmo2/creator/BlockCreator.java
mkouba: You're right, I missed the `CodeBuilder::ifnull` part in `If.Kind.EQ`. Will rewrite this to a default method.

src/main/java/io/quarkus/gizmo2/impl/BlockCreatorImpl.java
mkouba: Removed.

Ladicek:
> > > add BlockCreator#readStaticField() and BlockCreator#writeStaticField()
> > 
> > 
> > Given we currently have `BlockCreator.get()` and `set()`, how about we call these `getStatic()` and `setStatic()`? Seems more consistent.
> > I originally thought it would be enough to add `Expr.staticField()` to `BlockCreator`, but I agree this is better.
> 
> How `getStaticField()`/`setStaticField()`? `getStatic()`/`setStatic()` sounds too ambiguous to me...

Works for me, thanks!

src/main/java/io/quarkus/gizmo2/impl/BlockCreatorImpl.java
```diff
@@ -902,7 +901,7 @@ private void doIf(final Expr cond, final Consumer<BlockCreator> whenTrue, final
         }
         doIfInsn(CD_void, cond, wt, wf);
     }
-
+
```
Ladicek, line 904: Obligatory formatting nitpicking :-)

src/main/java/io/quarkus/gizmo2/impl/StaticFieldVarImpl.java
```diff
@@ -66,17 +66,18 @@ public String itemName() {
...
-                node = value.process(node, op);
                 if (mode != AccessMode.AsDeclared) {
                     node = ConstantImpl.ofStaticFieldVarHandle(desc).process(node.prev(), op);
+                } else {
+                    node = value.process(node.prev(), op);
                 }
                 return node;
             }
-
+
```
Ladicek, line 76: Obligatory formatting nitpicking :-)

Ladicek (approved):

src/main/java/io/quarkus/gizmo2/impl/BlockCreatorImpl.java
mkouba: Uf, we should really configure some formatting tool...

src/main/java/io/quarkus/gizmo2/impl/StaticFieldVarImpl.java
```diff
@@ -66,17 +66,18 @@ public String itemName() {
...
                 }
                 return node;
             }
-
+            
             public void writeCode(final CodeBuilder cb, final BlockCreatorImpl block) {
                 switch (mode) {
                     case AsDeclared -> {
-                        cb.putstatic(owner(), name(), type());
+                        cb.putstatic(owner(), name(), desc().type());
```
dmlloyd, line 80: I made this mistake in many places. 😞

dmlloyd (approved):

PR #285, opened 2025-04-03
Complete support for `try/catch/finally`

### LINKED ISSUE #233: Throw doesn't work correctly with try/catch
I added a note in the code, which reminded me that this still needs to be done. It's just a bit complex so I'll need a day for it.

dmlloyd:
Fixes #233


Ladicek (approved):
I honestly don't feel like reviewing this. I went through the code but didn't understand most of it. But test coverage seems good enough, so... +1 I guess :-)

dmlloyd:
I really should have broken it up into several monotonic commits, and in fact I would have done so if I'd known how I was going to solve the problem when I started.

Here's a summary of the conceptual things I changed:

- Changed some `BiConsumer<BlockCreator, Expr>` to `BiConsumer<BlockCreator, ? super LocalVar>`
- Added a `caughtName` to name the variable of the caught exception
- Made sure that division and remainder operations are always "bound" even if their inputs are not, because they can throw an exception (we could narrow this down to only div/rem on integer types) - the basic defect here is that an unused div/rem will get deleted otherwise, losing the side-effect of checking for division by zero (not that I recommend doing `n/x` as a test to see if `x` is zero, but correct is correct and incorrect is incorrect!)
- Extracted all anonymous subclasses of `Goto` to top-level classes
- Deleted some unused code (incl. the incomplete/nonfunctional `Cleanup` idea)
- Modify `Goto` so that the target label method accepts the "from" block to support `finally`
- Add new `CleanupKey` concept which is the primary interface for instructions which have to run cleanup on the way out of `finally` blocks (this includes the `Goto` subclasses plus `Return`, but not `Throw` (or method calls or other ops which may throw), which is always handled by a catch-all block)
- Separated `try` behaviors into two separate `Item` subtypes: one for `try`/`catch` and one for `try`/`finally`; `TryCreatorImpl` is no longer an `Item` and instead composes a combination of one or both of these subtypes after `try_` completes
- Update `TestClassMaker` so that `ClassFormatError` also triggers a nice class dump
- Testing

Ladicek:
I spotted and understood the first few described changes, but the gist of the issue, not really. I don't even know what "redo" is :-) Maybe you could walk us through this on our Monday call?

dmlloyd:
Sure, that works for me.

PR #288, opened 2025-04-07
Do not allow builders to accept raw `Expr`

### LINKED ISSUE #248: Gizmo2: Revisit generic arguments usage
We have several methods which accept `List<Expr>` or `Consumer<SomeCreator>`. These probably should be revised to be `List<? extends Expr>` and `Consumer<? super SomeCreator>`.

dmlloyd:
Fixes #248


Ladicek (approved):
I was wondering why `? super LocalVar`, but then figured it's a `LocalVar` that is passed to all the affected `BiConsumer`s and it's probably a good idea to highlight that it's freely reusable. LGTM

PR #297, opened 2025-04-07
Implement instance field initializers

### LINKED ISSUE #250: Gizmo2: Instance field initializers
There isn't great infrastructure for this at the moment, but it would be good if we could put initializers with their fields like we can do for static fields. The initializers would be rolled into every constructor right after the `super` call (this might be tricky to detect though).

dmlloyd:
Fixes #250


src/main/java/io/quarkus/gizmo2/creator/TypeCreator.java
```diff
@@ -128,7 +130,16 @@ default void implements_(Class<?> interface_) {
      *
      * @param builder the builder (must not be {@code null})
      */
-    void initializer(Consumer<BlockCreator> builder);
+    void staticInitializer(Consumer<BlockCreator> builder);
+
+    /**
+     * Add a general instance initializer block to the type.
+     * A type may have many static initializers;
```
Ladicek, line 137:
```suggestion
     * A type may have many instance initializers;
```

src/main/java/io/quarkus/gizmo2/impl/InstanceFieldCreatorImpl.java
```diff
@@ -3,16 +3,51 @@
...
+        initializer = (b0 -> {
+            FieldDesc desc = desc();
+            // todo: replace ThisExpr from a shared instance on TypeCreator
+            b0.set(new ThisExpr(owner()).field(desc), b0.blockExpr(desc.type(), init));
+        });
+    }
+
+    private void checkOneInit() {
+        if (initial != null || initializer != null) {
+            throw new IllegalStateException("A static field may have only one initializer");
```
Ladicek, line 39:
```suggestion
            throw new IllegalStateException("An instance field may have only one initializer");
```

src/test/java/io/quarkus/gizmo2/FieldAccessTest.java
```diff
@@ -110,4 +112,74 @@ public void testConstantField() {
...
+        TestClassMaker tcm = new TestClassMaker();
+        Gizmo g = Gizmo.create(tcm);
+        String className = "io.quarkus.gizmo2.InstanceInitializedFieldTest";
+        g.class_(className, zc -> {
+            // create an instance field and initialize it to a non-constant value
+            FieldDesc field1 = zc.field("field1", ifc -> {
+                ifc.withType(String.class);
+                ifc.final_();
+                ifc.withInitializer(b0 -> {
+                    b0.yield(b0.withString(Constant.of("Hello")).concat(Constant.of(" world")));
```
Ladicek, line 161: The `yield()` call here is not entirely intuitive, but I'm not sure what we can do about that. Maybe make `withInitializer()` take `Function<BlockCreator, Expr>` instead of `Consumer<BlockCreator>`?

Ladicek (approved):
LGTM baring the typos.

src/test/java/io/quarkus/gizmo2/FieldAccessTest.java
dmlloyd: This is not specific to this case... any block which has an output type (including switch expressions, `?:`, etc.) works this way. I agree it's not super intuitive and it can be easy to mess it up, but I don't see a good way of changing it in a uniform way without breaking a lot of things (or else changing *all* `Consumer` to `Function` and requiring users to return a `void` constant or something).

dmlloyd: We can have a separate discussion on that, at any rate.

src/main/java/io/quarkus/gizmo2/creator/FieldCreator.java
```diff
@@ -41,4 +45,72 @@ public sealed interface FieldCreator extends MemberCreator, SimpleTyped permits
...
+    default void withInitial(String initial) {
+        withInitial(initial == null ? Constant.ofNull(CD_String) : Constant.of(initial));
+    }
+
+    /**
+     * Provide an initial constant value for this field.
+     *
+     * @param initial the initial value
+     */
+    default void withInitial(Class<?> initial) {
```
mkouba, line 106: Hm, do you expect many `Class<?>` fields? :wink:

mkouba (commented):
A bit late but still - this is awesome!

There are too many `withInitial()` overloads for my taste but that's not a big deal.

PR #298, opened 2025-04-07
Non capturing lambdas should be singleton

### LINKED ISSUE #266: Gizmo2: Non-capturing lambdas should be singletons
If a lambda doesn't capture any values, it can be a singleton, which will improve runtime performance slightly. This could be done by returning a static getter instead of a constructor method handle in this case.

dmlloyd:
The existing tests already cover capturing and non-capturing cases. Also fixes a bug where a relational- or compare-to-zero-if statement at the end of a block is still considered to possibly fall through even if none of its branches fall through.

Fixes #266


Ladicek (approved):
I probably don't understand how the word "singleton" is used here, but this is about evaluating the lambda once and returning a MH to the result, instead of creating a MH that evaluates the lambda on each call?

dmlloyd:
It's about the lambda instance - the object which implements the functional interface. If there are captures, we must construct a new instance each time with the captured values. If not, we can return the same instance each time.

PR #303, opened 2025-04-08
tests annotations and invocations + refactor flag setting

Ladicek:


Ladicek:
I've got a few questions here:

- in `Annotatable`, we default an unset retention to `RUNTIME`, even though the actual default is specified to be `CLASS` -- I guess that makes sense, so I documented that, but I wonder if that wasn't an oversight and we actually want to use the specified default instead
- in `AnnotationCreator`, I named all the array-accepting methods `withArray()`, just to avoid the pathological case of single-element arrays accidentally adding a non-array annotation value instead, WDYT?
- in `AnnotationCreator`, I made 2 `withArray()` methods take a `List` instead of a varargs because interface methods cannot be made `@SafeVarargs`, WDYT?

dmlloyd (approved):

dmlloyd:
> I've got a few questions here:
> 
> * in `Annotatable`, we default an unset retention to `RUNTIME`, even though the actual default is specified to be `CLASS` -- I guess that makes sense, so I documented that, but I wonder if that wasn't an oversight and we actually want to use the specified default instead

Either way is fine for me. I think a default of `RUNTIME` is more common, but going with the normal default is arguably a better policy, so do whatever you think is best.

> * in `AnnotationCreator`, I named all the array-accepting methods `withArray()`, just to avoid the pathological case of single-element arrays accidentally adding a non-array annotation value instead, WDYT?

Makes sense.

> * in `AnnotationCreator`, I made 2 `withArray()` methods take a `List` instead of a varargs because interface methods cannot be made `@SafeVarargs`, WDYT?

Fine by me.

PR #305, opened 2025-04-09
fix unboxing primitive values and unboxing via casting

Ladicek:
Boxing already boxed values and boxing via casting used to work properly, but unboxing already unboxed values and unboxing via casting were buggy. This commit fixes those bugs, adds tests and makes the box and unbox code paths more similar.

Additionally, this commit adds an explicit check for attempting to box/unbox `void`, which is never a valid operation.

### DIFF
--- a/src/main/java/io/quarkus/gizmo2/impl/BlockCreatorImpl.java
+++ b/src/main/java/io/quarkus/gizmo2/impl/BlockCreatorImpl.java
@@ -3,6 +3,7 @@ package io.quarkus.gizmo2.impl;
 import static io.quarkus.gizmo2.impl.Preconditions.requireArray;
 import static io.quarkus.gizmo2.impl.Preconditions.requireSameLoadableTypeKind;
 import static io.quarkus.gizmo2.impl.Preconditions.requireSameTypeKind;
+import static io.smallrye.common.constraint.Assert.impossibleSwitchCase;
 import static java.lang.constant.ConstantDescs.*;
 import static java.util.Collections.*;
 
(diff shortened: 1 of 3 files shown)


dmlloyd (approved):

src/main/java/io/quarkus/gizmo2/impl/BlockCreatorImpl.java
```diff
@@ -269,20 +270,17 @@ public void remAssign(final LValueExpr var, final Expr arg) {
...
+    private static final Map<ClassDesc, ClassDesc> boxTypes = Map.of(
+        CD_boolean, CD_Boolean,
+        CD_byte, CD_Byte,
+        CD_char, CD_Character,
+        CD_short, CD_Short,
+        CD_int, CD_Integer,
+        CD_long, CD_Long,
+        CD_float, CD_Float,
+        CD_double, CD_Double,
+        CD_void, CD_Void
```
mkouba, line 282: If you can't box/unbox `void` then why this map entry?

Ladicek: That is a great question that I have asked myself as well, and frankly, I'm not sure. It's been there, so I kept it :-) I guess there might eventually be a usecase that would require it, but that's a great case of YAGNI...

Ladicek:
Rebased on `main` and formatted.

src/main/java/io/quarkus/gizmo2/impl/BlockCreatorImpl.java
dmlloyd: We would use it when autoboxing return types, probably. I think it was being used at one point for something else, maybe? But you can remove it if you want; we can always add it back if needed.

I would suggest that the second map be built from the first map though, rather than duplicating the map entries.

Ladicek: Makes sense, will do.

Ladicek: Done. I replaced the second map by a call to `reverseMap()`. I've kept the `void`/`Void` types in the map just to avoid possible future problems when we have to add it. It's there since the beginning, so everyone has to deal with it.

dmlloyd (approved):

PR #307, opened 2025-04-09
improve javadoc and formatting

### LINKED ISSUE #306: automated formatting and import sorting
We really should have automated formatting and import sorting, to preserve some basic code consistency. I'm thinking using the Quarkus configuration, as in Gizmo 1, is probably a decent starting point, even if I don't particularly like it :-)

Ladicek:
Resolves #306

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -41,10 +41,13 @@
         <maven.compiler.release>17</maven.compiler.release>
 
         <version.classfile>24.0</version.classfile>
-        <version.smallrye-common>2.11.0</version.smallrye-common>
         <version.junit>5.12.1</version.junit>
+        <version.smallrye-common>2.11.0</version.smallrye-common>
 
+        <version.formatter-maven-plugin>2.26.0</version.formatter-maven-plugin>
+        <version.impsort-maven-plugin>1.12.0</version.impsort-maven-plugin>
         <version.nexus-staging-plugin>1.7.0</version.nexus-staging-plugin>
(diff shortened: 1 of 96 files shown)


Ladicek:
Commits best reviewed in isolation.

Note that this PR conflicts with #305, so the one that isn't merged first will have to be rebased.

gsmet:
Merge the formatting PR, first, it's going to be easy to fix the other one.

BTW, you probably need something similar to the following in your workflow:

https://github.com/quarkiverse/quarkus-github-app/blob/cb6760716d8dca737a7d27de136a5ca2c102d4e6/.github/workflows/build.yml#L62-L63C19

so that you validate the formatting, otherwise it will reformat things and not complain about it.

dmlloyd (approved):

Ladicek:
> BTW, you probably need something similar to the following in your workflow:
> 
> https://github.com/quarkiverse/quarkus-github-app/blob/cb6760716d8dca737a7d27de136a5ca2c102d4e6/.github/workflows/build.yml#L62-L63C19
> 
> so that you validate the formatting, otherwise it will reformat things and not complain about it.

I think we should have this already, because:

1. The GitHub Actions CI build runs `mvn -B clean package -Dno-format`
2. The Maven profile activated with `-Dno-format` runs the `validate` goal of `formatter-maven-plugin` (as well as the `check` goal of `impsort-maven-plugin`).

So we should be good there.

PR #310, opened 2025-04-09
Forbid duplicate methods/fields/ctors

### LINKED ISSUE #300: verify duplicate methods/fields
I've ran at least once into the case where Gizmo 2 is happy to forward to the ClassFile API a class that has 2 conflicting methods (the same name and signature), and the ClassFile API has happily generated a `.class` file that failed to load. I didn't try with fields, but I guess it's the same situation.

We should check this in Gizmo.

dmlloyd:
Fixes #300


Ladicek (approved):
LGTM, but it would be good to have a comment on `fields`/`methods` that the boolean value means whether the given member is `static`. Took me a bit to figure that out :-)

Ladicek (approved):

PR #311, opened 2025-04-09
add tests for arithmetic operations and fix type check

Ladicek:
The type check in `BinOp` used to assume that all binary operations accept operands of the same type. That is not true: the shifts (shl, shr, ushr) always accept an `int` second operand, even if the first operand is `long`. This issue is fixed now.

Further, this commit adds tests for arithmetic operations on `int`, `long` and `double`.


Ladicek:
I didn't add tests for other type only out of pure laziness: I would also have to define functional interfaces for those types :-)

PR #319, opened 2025-04-10
Release 2.0.0.Alpha3

gastaldi:


gsmet:
@dmlloyd note that this PR will change the version in `main` to `2.999-SNAPSHOT` - mostly because otherwise updating the next version will be a pain. It could also be changed to `999-SNAPSHOT` if you prefer - it won't conflict with Gizmo 1 even if we push this infra to Gizmo 1 too given they have different GAs.

Let us know what you think.

dmlloyd:
It isn't that painful to use the next stable version (in this case `2.0.0-SNAPSHOT`), we've done this plenty of times in the past. What problem is being solved here?

gastaldi:
I'm fine with either, I guess @gsmet concerns' is because once 2.1.x is released you need to change that too, whereas if you use 2.999-SNAPSHOT you don't ever touch that variable

gsmet:
Exactly that ^

But the idea is that you could trigger the releases yourself from now on so we can do whatever you prefer :)

gastaldi:
I reverted to 2.0.0-SNAPSHOT for now, we can discuss about the 999 idea in a later stage

dmlloyd (approved):

PR #330, opened 2025-04-14
hide some ClassFile API

### LINKED ISSUE #331: avoid leaking the ClassFile API for annotation creation
### Issue description

_No response_

### LINKED ISSUE #332: avoid leaking the ClassFile API for specifying class version
### Issue description

_No response_

Ladicek:
- the API to create annotations no longer leaks the ClassFile API
- add the `ClassVersion` enum

Fixes #331
Fixes #332

### DIFF
--- a/src/main/java/io/quarkus/gizmo2/impl/AnnotationCreatorImpl.java
+++ b/src/main/java/io/quarkus/gizmo2/impl/AnnotationCreatorImpl.java
@@ -1,0 +1,519 @@
+package io.quarkus.gizmo2.impl;
+
+import static io.smallrye.common.constraint.Assert.checkNotNullParam;
+
+import java.lang.annotation.Annotation;
+import java.lang.constant.ClassDesc;
+import java.lang.invoke.MethodHandleInfo;
+import java.lang.invoke.SerializedLambda;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.function.Consumer;
+
+import io.github.dmlloyd.classfile.AnnotationElement;
+import io.github.dmlloyd.classfile.AnnotationValue;
+import io.quarkus.gizmo2.creator.AnnotationCreator;
+
+public final class AnnotationCreatorImpl<A extends Annotation> implements AnnotationCreator<A> {
+    /**
+     * Factory for creating annotation instances.
+     *
+     * @param type the annotation class (must not be {@code null})
+     * @param builder the builder for the annotation contents (must not be {@code null})
+     * @return the constructed annotation (not {@code null})
+     * @param <A> the annotation type
+     */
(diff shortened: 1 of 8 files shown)


Ladicek:
Probably best to review commits in isolation, even though the second one is quite small.

src/main/java/io/quarkus/gizmo2/impl/AnnotationCreatorImpl.java
```diff
@@ -0,0 +1,528 @@
...
+        String sig = Util.serializedLambda(prop).getImplMethodSignature();
+        assert sig.startsWith("()L") && sig.endsWith(";");
+        String annotationClass = sig.substring(3, sig.length() - 1).replace('/', '.');
+        Class<S> clazz;
+        try {
+            ClassLoader cl = Thread.currentThread().getContextClassLoader();
+            if (cl == null) {
+                cl = AnnotationCreator.class.getClassLoader();
+            }
+            clazz = (Class<S>) Class.forName(annotationClass, false, cl);
```
dmlloyd, line 236: I think we should consider requiring that a `Class<S>` be passed in instead of trying to do class loader magic. TCCL is almost certainly the wrong one to use at any rate.

dmlloyd: Ideally yes, but in practice I can't think of a totally safe/totally correct way to do it automatically. The lambda instance does not seem to have a concrete generic type, so we can't probe it that way. AFAICT we can't know for sure what class loader originated the type of the nested annotation.

dmlloyd: Actually, I'm being silly. We can examine the method on the enclosing annotation type.

src/main/java/io/quarkus/gizmo2/impl/TypeCreatorImpl.java
```diff
@@ -83,8 +85,15 @@ public ClassOutputImpl output() {
...
-        this.version = Objects.requireNonNull(version, "version");
+    @Override
+    public void withVersion(final Runtime.Version version) {
+        checkNotNullParam("version", version);
+        this.version = ClassFileFormatVersion.valueOf(version);
+    }
+
+    public void withVersion(final ClassVersion version) {
+        checkNotNullParam("version", version);
+        this.version = ClassFileFormatVersion.valueOf("RELEASE_" + version.number());
```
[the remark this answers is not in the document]
Ladicek: Done.

src/main/java/io/quarkus/gizmo2/impl/AnnotationCreatorImpl.java
Ladicek: Done.

PR #336, opened 2025-04-14
make Var an LValueExpr

Ladicek:
Now that `This` is no longer a `Var`, it makes sense for `Var` to simply be a subinterface of `LValueExpr`. This improves clarity of the API.

Related to #281


dmlloyd:
Since all assignable things are "variables" according to JLS (including array dereferences 🤯), what if we merged `LValueExpr`/`Assignable` and `Var` into one interface and call it `Var`?

dmlloyd (approved):

Ladicek:
Yeah agree, I think array dereference is the prime example of an `LValueExpr` that is not a `Var`. I'm not sure if there are more, possibly not.

Issue #222, opened 2025-03-11
Gizmo2: consider the default flags for classes and members


mkouba:
I agree that the defaults should just follow the JLS. But in Gizmo1 the defaults differ. For example, the `MethodCreator` produces `public` methods by default, fields are `private` (unless declared on an interface), etc. In some cases, it's more convenient. By the way, even in Gizmo2 the generated class is `public` by default, and not package private.

I wonder if we should stick with the same rules to ease the migration, or maybe introduce something like "default flag profiles"?

```java
public interface AccessFlagProfile {
    public Collection<AccessFlag> methodFlags();
}

public class PublicFinalMethods implements AccessFlagProfile {
   public Collection<AccessFlag> methodFlags() {
      return List.of(AccessFlag.PUBLIC, AccessFlag.FINAL);
   }
}

g.class_(ClassDesc.of("org.acme.Foo"), cc -> {
   cc.withAccessFlagProfile(new PublicFinalMethods()); // all methods will be "public final" by default
}
```

And have some predefined profiles... like `GIZMO1`.

Maybe I'm wrong. I'm just throwing ideas...

dmlloyd:
Maybe we could set the defaults at a `ClassOutput` level like this:

```java
co.defaultAccess(ALL_TYPES, PUBLIC); // all types would be public by default
co.defaultAccess(CLASSES, PACKAGE_PRIVATE); // ...except classes which would be package-private by default
co.defaultAccess(CLASSES, ABSTRACT_METHODS, PROTECTED); // abstract methods on classes are protected
// ...etc
```

mkouba:
> Maybe we could set the defaults at a `ClassOutput` level like this:

Yes, that's exactly what I had in mind. However, I'd prefer to define a single interface + builder-like API, something like:

```java
co.defaultAccess(DefaultAccess.builder()
   .all(PUBLIC)
   .classes(PACKAGE_PRIVATE)
   .abstractMethods(PROTECTED)
   .build()
);

interface DefaultAccess {
   default List<AccessFlag> forClass(ClassDesc clazz) {
      return List.of(PUBLIC);  
   }
  default List<AccessFlag> forInterface(ClassDesc clazz) {
     return List.of(PUBLIC);  
  }
  default List<AccessFlag> forMethod(MethodDesc method) {
     return List.of(PUBLIC);
  }
  default List<AccessFlag> forStaticMethod(MethodDesc method) {
     return forMethod(method);
  }
[...]

PR #338, opened 2025-04-14
Change organization of `Gizmo` and `ClassOutput`

dmlloyd:
This more closely aligns with Gizmo 1, and simplifies the API for class creation, but still leaves room to implement #222 in some sane way.

It stops short of using static methods, but with static methods it becomes difficult to establish a configuration that would persist but have constrained scope, like what would be required for #222.


dmlloyd:
An alternative approach to this would be to use static methods, but also require some kind of configuration to be passed in. In current Quarkus usages, we'd have to make that configuration readily available.

A non-alternative is to have a mutable global configuration.

src/main/java/io/quarkus/gizmo2/impl/GizmoImpl.java
```diff
@@ -1,147 +1,53 @@
...
-
-    public StaticFinalFieldConstant staticFinalFieldConstant(final FieldDesc desc) {
-        return new StaticFinalFieldConstant(desc);
-    }
 
-    public ClassOutput classOutput(final BiConsumer<ClassDesc, byte[]> outputHandler) {
-        return new ClassOutputImpl(this, outputHandler);
+        ClassFile cf = ClassFile.of(ClassFile.StackMapsOption.GENERATE_STACK_MAPS);
+        byte[] bytes = cf.build(desc, zb -> {
+            zb.withVersion(ClassFile.JAVA_17_VERSION, 0);
```
Ladicek, line 30: Do we need this? It's gonna be overwritten by `preAccept`. If we do need this, maybe there should be a comment that the value is not gonna survive?

Ladicek (approved):
LGTM! Not merging just to wait for an answer on the one given question.

Ladicek:
I think this is great. I agree we don't need `static` methods on `Gizmo`, the main value is that `ClassOutput` works a lot closer to Gizmo 1.

PR #344, opened 2025-04-15
Repeat annotation validation

### LINKED ISSUE #341: Disallow repeat annotations (unless `@Repeatable`)
### Issue description

We should prevent annotations from being added multiple times. We should also automatically handle annotations which are `@Repeatable` so the user does not need to think too much about it.

dmlloyd:
Also, support for `@Repeatable`. Fixes #341

### DIFF
--- a/src/main/java/io/quarkus/gizmo2/Annotatable.java
+++ b/src/main/java/io/quarkus/gizmo2/Annotatable.java
@@ -3,6 +3,7 @@ package io.quarkus.gizmo2;
 import static io.smallrye.common.constraint.Assert.checkNotNullParam;
 
 import java.lang.annotation.Annotation;
+import java.lang.annotation.Repeatable;
 import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import java.lang.constant.ClassDesc;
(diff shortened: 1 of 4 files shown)


src/main/java/io/quarkus/gizmo2/Annotatable.java
```diff
@@ -22,8 +23,14 @@ public sealed interface Annotatable permits MemberCreator, ParamCreator, TypeCre
      * Add an annotation with no elements. If {@code annotationClass} has no {@link Retention}
      * annotation, {@link RetentionPolicy#RUNTIME} is assumed. This is the most commonly used
      * value, even though {@code CLASS} is the specified default.
+     * <p>
+     * If the annotation is {@linkplain Repeatable repeatable}, and adding this annotation would
+     * cause there to be more than one instance of the given annotation on this element,
+     * then the existing annotation and the given annotation are automatically wrapped in the
+     * repeatable annotation wrapper.
```
Ladicek, line 30:
```suggestion
     * container annotation.
```

it I believe the proper wording.

Ladicek (approved):
LGTM, but we should have a test :-)

dmlloyd:
Sure enough, the test found a bug. :)

PR #353, opened 2025-04-16
small improvements

Ladicek:
- add `MethodTyped.parameterCount()`
- add `BlockCreator.newEmptyArray()` overloads that take an `int` size
- add `BlockCreator.throw_()` overloads that take an `Expr` message
- add `ConstructorDesc.of()` overloads that accept vararg parameter types
- validate the return type on creation of `ConstructorDesc`
- add `TestClassMaker.loadClass()`
- improve the exception message when an `Item` is used on an unexpected place

### DIFF
--- a/src/main/java/io/quarkus/gizmo2/creator/BlockCreator.java
+++ b/src/main/java/io/quarkus/gizmo2/creator/BlockCreator.java
@@ -381,7 +381,7 @@ public sealed interface BlockCreator extends SimpleTyped permits BlockCreatorImp
     // arrays
 
     /**
-     * Create a new, empty array of the given type.
+     * Create a new, empty array of the given type with given size.
      *
      * @param componentType the component type (must not be {@code null})
      * @param size the size of the array (must not be {@code null})
@@ -390,7 +390,18 @@ public sealed interface BlockCreator extends SimpleTyped permits BlockCreatorImp
     Expr newEmptyArray(ClassDesc componentType, Expr size);
 
     /**
-     * Create a new, empty array of the given type.
+     * Create a new, empty array of the given type with given size.
+     *
+     * @param componentType the component type (must not be {@code null})
+     * @param size the size of the array
+     * @return the expression for the new array (not {@code null})
+     */
+    default Expr newEmptyArray(ClassDesc componentType, int size) {
+        return newEmptyArray(componentType, Constant.of(size));
+    }
+
+    /**
+     * Create a new, empty array of the given type with given size.
      *
      * @param componentType the component type (must not be {@code null})
      * @param size the size of the array (must not be {@code null})
@@ -398,6 +409,17 @@ public sealed interface BlockCreator extends SimpleTyped permits BlockCreatorImp
      */
     default Expr newEmptyArray(Class<?> componentType, Expr size) {
         return newEmptyArray(Util.classDesc(componentType), size);
+    }
+
+    /**
+     * Create a new, empty array of the given type with given size.
+     *
+     * @param componentType the component type (must not be {@code null})
+     * @param size the size of the array
+     * @return the expression for the new array (not {@code null})
+     */
+    default Expr newEmptyArray(Class<?> componentType, int size) {
+        return newEmptyArray(Util.classDesc(componentType), Constant.of(size));
     }
 
     /**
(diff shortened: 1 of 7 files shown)


Ladicek:
Commits are best reviewed in isolation, although they are all pretty small. The last commit adds certain (nontrivial?) overhead, so I'd be willing to gate this behind a system property, but I found it tremendously helpful when figuring out what `Expr` exactly needs to be converted to a local variable.

src/main/java/io/quarkus/gizmo2/creator/BlockCreator.java
```diff
@@ -2539,7 +2571,7 @@ default void throw_(Class<? extends Throwable> type) {
     /**
      * Throw a new exception of the given type with a message.
      *
-     * @param type the exception type (must not be {@code null})
```
dmlloyd, line 2542: I think this edit hit the wrong line.

src/main/java/io/quarkus/gizmo2/creator/BlockCreator.java
```diff
@@ -2550,6 +2582,20 @@ default void throw_(Class<? extends Throwable> type, String message) {
         }
     }
 
+    /**
+     * Throw a new exception of the given type with a message.
+     *
+     * @param type the exception type
```
dmlloyd, line 2588: Same here; `type` is not nullable but `message` is nullable.

src/main/java/io/quarkus/gizmo2/desc/ConstructorDesc.java
```diff
@@ -37,7 +37,33 @@ static ConstructorDesc of(ClassDesc owner) {
...
-     * @param owner the descriptor of the class which contains the member (must not be {@code null})
+     * @param owner the descriptor of the class which contains the constructor (must not be {@code null})
+     * @param paramTypes a list of parameter types (must not be {@code null})
+     * @return the constructor descriptor (not {@code null})
+     */
+    static ConstructorDesc of(ClassDesc owner, Class<?>... paramTypes) {
+        ClassDesc[] params = new ClassDesc[paramTypes.length];
+        for (int i = 0; i < paramTypes.length; i++) {
+            params[i] = Util.classDesc(paramTypes[i]);
+        }
```
dmlloyd, line 48: Maybe `Stream.of(paramTypes).map(Util::classDesc).toArray(ClassDesc[]::new)`?

src/main/java/io/quarkus/gizmo2/impl/Item.java
```diff
@@ -18,6 +18,7 @@
 import io.quarkus.gizmo2.impl.constant.ConstantImpl;
 
 public abstract non-sealed class Item implements Expr {
+    private final String creationSite = Util.callerOutsideGizmo();
```
dmlloyd, line 21: I wonder if this might have a significant performance impact. We might want to disable it by default.

src/main/java/io/quarkus/gizmo2/creator/BlockCreator.java
Ladicek: Ouch, you're right, my bad. Will fix.

src/main/java/io/quarkus/gizmo2/desc/ConstructorDesc.java
Ladicek: Will do.

src/main/java/io/quarkus/gizmo2/impl/Item.java
Ladicek: Right, I was expecting this comment :-) I have no idea how I would quantify the impact, but the number of `Item`s we allocate is nontrivial, so the impact here is likely to be nontrivial as well. I'll add a system property.

src/main/java/io/quarkus/gizmo2/creator/BlockCreator.java
Ladicek: Done.

src/main/java/io/quarkus/gizmo2/creator/BlockCreator.java
Ladicek: Done.

src/main/java/io/quarkus/gizmo2/desc/ConstructorDesc.java
Ladicek: Done.

Ladicek:
In addition to addressing the comments, I reworked the commit that updates `TestClassMaker`. The class has gained a method `forClass()` that returns a helper object of class `TestClassOps` with the same methods as `TestClassMaker`, except for the given class (and not for the last accepted class). Actually, the helper methods directly on `TestClassMaker` now delegate to `TestClassOps` created for the last accepted class. That feels more natural and more universal.

PR #358, opened 2025-04-22
`unsafeCast` -> `uncheckedCast`

dmlloyd:
On further thought I've decided that `uncheckedCast` is a better name. Everyone knows what "unchecked" means - it means the cast isn't checked. The cast is not "unsafe" in any way - if the cast is valid, it's safe, and if it's invalid, it will fail verification. It also intuitively complements the `CHECKCAST` instruction.


Ladicek:
To me, "unchecked cast" sounds like "unchecked conversion" (per https://docs.oracle.com/javase/specs/jls/se17/html/jls-5.html#jls-5.1.9), which is a totally different thing. An "unsafe cast" is not type safe, so it _is_ unsafe in one specific meaning of the word. But I'm not too stubborn about this :-)

PR #359, opened 2025-04-23
Cleanups and unused method removals

dmlloyd:


Ladicek (approved):

src/main/java/io/quarkus/gizmo2/impl/Util.java
```diff
@@ -200,9 +157,9 @@ public static <T, R> List<R> reinterpretCast(List<T> list) {
     }
 
     public static <K, V> Map<V, K> reverseMap(Map<K, V> original) {
-        Map<V, K> result = new HashMap<>();
-        original.forEach((k, v) -> result.put(v, k));
-        return result;
+        return original.entrySet().stream().collect(Collectors.toUnmodifiableMap(
+                Map.Entry::getValue,
+                Map.Entry::getKey));
```
Ladicek, line 162: This seems marginally harder to read than the original :-), but I see what you mean here and it's OK.

PR #363, opened 2025-04-24
Atomic operations support

dmlloyd:
`VarHandle` is hard to use, and even harder to use from a bytecode generator. We can make it much easier by providing atomic operations which operate on any valid `Assignable`.

This is better than providing a `withVarHandle()` ops class because each kind of assignable has a different shape and bootstrap strategy. We can hide all of that fairly easily.

Also fix a problem where field types are not auto-set when defining them with an initializer.


src/main/java/io/quarkus/gizmo2/creator/BlockCreator.java
```diff
@@ -398,6 +404,392 @@ default void dec(Assignable var) {
...
+     * @param other the value to bitwise-or with the target (must not be {@code null})
+     * @return the previous value of the target expression (not {@code null})
+     * @throws IllegalArgumentException if the target variable cannot be accessed atomically
+     */
+    default Expr getAndBitwiseOr(Assignable var, Expr other) {
+        return getAndBitwiseOr(var, other, MemoryOrder.Volatile);
+    }
+
+    /**
+     * Atomically get, bitwise-or, and store the value of the target assignable expression.
```
Ladicek, line 667:
```suggestion
     * Atomically get, bitwise-and, and store the value of the target assignable expression.
```

src/main/java/io/quarkus/gizmo2/creator/BlockCreator.java
```diff
@@ -398,6 +404,392 @@ default void dec(Assignable var) {
...
+        } else if (var instanceof Item.FieldDeref fd) {
+            varHandle = Const.ofFieldVarHandle(fd.desc());
+            args = List.of(fd.instance(), expected, update);
+            typeDesc = MethodTypeDesc.of(var.type(), fd.instance().type(), var.type(), var.type());
+        } else if (var instanceof Item.ArrayDeref ad) {
+            varHandle = Const.ofArrayVarHandle(ad.type().arrayType());
+            args = List.of(ad.array(), ad.index(), expected, update);
+            typeDesc = MethodTypeDesc.of(var.type(), ad.array().type(), CD_int, var.type(), var.type());
+        } else {
+            throw new IllegalArgumentException("Unsupported target for atomic operations");
```
Ladicek, line 510:
```suggestion
            throw new IllegalArgumentException("Unsupported target for atomic operations: " + var);
```

src/main/java/io/quarkus/gizmo2/creator/BlockCreator.java
```diff
@@ -398,6 +404,392 @@ default void dec(Assignable var) {
...
+        } else if (var instanceof Item.FieldDeref fd) {
+            varHandle = Const.ofFieldVarHandle(fd.desc());
+            args = List.of(fd.instance(), arg);
+            typeDesc = MethodTypeDesc.of(var.type(), fd.instance().type(), var.type());
+        } else if (var instanceof Item.ArrayDeref ad) {
+            varHandle = Const.ofArrayVarHandle(ad.type().arrayType());
+            args = List.of(ad.array(), ad.index(), arg);
+            typeDesc = MethodTypeDesc.of(var.type(), ad.array().type(), CD_int, var.type());
+        } else {
+            throw new IllegalArgumentException("Unsupported target for atomic operations");
```
Ladicek, line 544:
```suggestion
            throw new IllegalArgumentException("Unsupported target for atomic operations: " + var);
```

Ladicek:
LGTM, just a few tiny suggestions.

Ladicek (approved):

PR #367, opened 2025-04-25
Introduce list/set/map constant support

dmlloyd:
Introduce two features:

* `List`/`Set`/`Map` plain constants (good for less than ~30 elements)
* External resource-backed constants for `List`/`Set`/`Map` of `String` specifically (no other constant types)

The constants are implemented in terms of condy.

The plain constants are a bit greedy with the constant pool. The amount of slots will amortize towards one slot per entry if enough are used, but for smaller collections it could be 2-3 slots per entry. Nevertheless these could be quite handy in certain circumstances where you have a small fixed set of, say, classes that you only use in one or two places and you don't want to go through the ceremony of setting up a field for it.

The resource-backed collections are unbounded in maximum size, however they will take up several constant pool slots (amortizing to 1 per collection if enough are used in one file).

### DIFF
--- a/src/main/java/io/quarkus/gizmo2/creator/ops/StringBuilderOps.java
+++ b/src/main/java/io/quarkus/gizmo2/creator/ops/StringBuilderOps.java
@@ -56,7 +56,7 @@ public final class StringBuilderOps extends ObjectOps implements ComparableOps {
      * @return this instance
      */
     public StringBuilderOps append(final Expr expr) {
-        switch (expr.type().descriptorString()) {
+        return new StringBuilderOps(bc, switch (expr.type().descriptorString()) {
             case "Z" -> invokeInstance(StringBuilder.class, "append", boolean.class, expr);
             case "B", "S", "I" -> invokeInstance(StringBuilder.class, "append", int.class, expr);
             case "J" -> invokeInstance(StringBuilder.class, "append", long.class, expr);
@@ -67,8 +67,7 @@ public final class StringBuilderOps extends ObjectOps implements ComparableOps {
             case "Ljava/lang/String;" -> invokeInstance(StringBuilder.class, "append", String.class, expr);
             case "Ljava/lang/CharSequence;" -> invokeInstance(StringBuilder.class, "append", CharSequence.class, expr);
             default -> invokeInstance(StringBuilder.class, "append", Object.class, expr);
-        }
-        return this;
+        });
     }
 
     /**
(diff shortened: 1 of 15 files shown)


src/main/java/io/quarkus/gizmo2/creator/ops/StringBuilderOps.java
```diff
@@ -67,8 +67,7 @@ public StringBuilderOps append(final Expr expr) {
             case "Ljava/lang/String;" -> invokeInstance(StringBuilder.class, "append", String.class, expr);
             case "Ljava/lang/CharSequence;" -> invokeInstance(StringBuilder.class, "append", CharSequence.class, expr);
             default -> invokeInstance(StringBuilder.class, "append", Object.class, expr);
-        }
-        return this;
+        });
```
Ladicek, line 70: Any reason for no longer returning `this`? It probably doesn't make a difference, but the method is literally documented to `@return this instance`.

dmlloyd: Yes, it fails if the wrapper object was created from an arbitrary expression since `this` would represent an object which is no longer on the stack.

The wrapper objects are set up to only be used once. The `StringBuilderOps` is a sort of weird exception in that there is special code to create a variable for it, but it's not guaranteed that its receiver will be in a variable.

Ladicek:
>The plain constants are a bit greedy with the constant pool. The amount of slots will amortize towards one slot per entry if enough are used, but for smaller collections it could be 2-3 slots per entry. Nevertheless these could be quite handy in certain circumstances where you have a small fixed set of, say, classes that you only use in one or two places and you don't want to go through the ceremony of setting up a field for it.
>
> The resource-backed collections are unbounded in maximum size, however they will take up several constant pool slots (amortizing to 1 or 2 per collection if enough are used in one file).

In other words, this means there's constant pool overhead in each class that uses this mechanism, due to condy, but it's fixed regardless of how many collections/elements are used. Right?

Ladicek:
Looks reasonable on the first sight.

dmlloyd:
> > The plain constants are a bit greedy with the constant pool. The amount of slots will amortize towards one slot per entry if enough are used, but for smaller collections it could be 2-3 slots per entry. Nevertheless these could be quite handy in certain circumstances where you have a small fixed set of, say, classes that you only use in one or two places and you don't want to go through the ceremony of setting up a field for it.
> > The resource-backed collections are unbounded in maximum size, however they will take up several constant pool slots (amortizing to 1 or 2 per collection if enough are used in one file).
> 
> In other words, this means there's constant pool overhead in each class that uses this mechanism, due to condy, but it's fixed regardless of how many collections/elements are used. Right?

More or less. Each collection needs one constant pool entry for the collection itself (as of my most recent code). A certain amount of constant pool entries are used by the bootstrap method(s), but those only need to be generated once (per collection type) even if many collections are used in the same class. That's basically it.

src/main/java/io/quarkus/gizmo2/impl/TypeCreatorImpl.java
```diff
@@ -60,7 +99,7 @@ public abstract sealed class TypeCreatorImpl extends AnnotatableCreatorImpl impl
     List<Consumer<BlockCreator>> postInits = List.of();
     private List<Signature.ClassTypeSig> interfaceSigs = List.of();
     private int flags;
-    private boolean hasLambdaBootstrap;
+    private int bootstraps;
```
Ladicek, line 102: I think an `EnumSet` would work the same here, but we can keep it manual.

Ladicek (approved):
Can't say I reviewed the condy bootstraps in detail, but overall LGTM.

PR #368, opened 2025-04-28
add Const.ofDefault()

Ladicek:


gsmet:
Looks like a wish come true. Thanks!

src/main/java/io/quarkus/gizmo2/Const.java
```diff
@@ -68,6 +71,40 @@ static Const ofNull(Class<?> type) {
...
     }
 
+    /**
+     * {@return a constant with the default value of given {@code type}}
+     * This is zero for primitive types and {@code null} for reference types.
+     *
+     * @param type the type of the constant (must not be {@code null})
+     */
+    static Const ofDefault(ClassDesc type) {
+        if (type.isPrimitive()) {
```
dmlloyd, line 81: I think you don't even need this "if"; you can make `ofNull` be the default case of the switch on the descriptor string and simplify this code a little bit.

Ladicek: Ah, that's a good point. Will do.

Ladicek: Done.

dmlloyd (approved):

PR #373, opened 2025-05-12
small improvements

Ladicek:


Ladicek:
Best to review each commit in isolation.

src/main/java/io/quarkus/gizmo2/impl/BlockCreatorImpl.java
```diff
@@ -682,13 +682,13 @@ public Expr cast(final Expr a, final ClassDesc toType) {
 
     public Expr uncheckedCast(final Expr a, final ClassDesc toType) {
         if (a.type().isPrimitive()) {
-            throw new IllegalArgumentException("Only object types may be unsafely cast");
+            throw new IllegalArgumentException("Only object types may be unchecked cast");
```
dmlloyd, line 677: "unchecked" is an adjective, not an adverb, so you'd have to say something like "Unchecked cast may only apply to object types".

src/main/java/io/quarkus/gizmo2/MethodTyped.java
```diff
@@ -52,6 +53,13 @@ default boolean isPrimitiveReturn() {
         return returnType().isPrimitive();
     }
 
+    /**
+     * {@return the list of parameter types}
+     */
+    default List<ClassDesc> parameterList() {
```
dmlloyd, line 60: These methods probably should be called `parameterTypes()`; they are not a list of parameters but their types.

src/main/java/io/quarkus/gizmo2/impl/Item.java
```diff
@@ -94,6 +94,10 @@ public Node pop(Node node) {
         }
     }
 
+    protected Node insertPopBefore(Node node) {
```
dmlloyd, line 97: This method, if I understand the code correctly, should likely be named something like `popInsertionPoint`. However, I am not certain that this algorithm is correct, because when iterating forwards, you cannot know what any given value is going to be used for, so I am not sure it is possible to reliably locate the values which will end up being used in the array or constructor call.

Even if you *can* reliably find the correct insertion point here, that information could be better used to recursively pop all of the arguments and then remove the original nodes.

I wonder if it would be a better solution to add an item at the end representing the complete operation, which has dependencies on each of its component operations. For array creation, it would have a dependency on each (`void`-typed) array store plus the initial array creation, and for constructor invocation, it would have a dependency on the `New` and `Invoke` items. Then it would be a relatively simple matter to implement `pop` on those nodes, since it could recursively delete/pop the constituent nodes, or else provide a correct location to insert the `Pop`.

Since `void`-type items cannot logically be popped, the custom `pop` operation for array creation would have to remove its own node, and then for each array store, call `remove` on each `ArrayStore` node and then immediately `pop` on each of its dependencies (this will delete the `Dup` and may delete the value being stored or may pop it if it has side-effects).

The custom `pop` for regular object creation would be a little simpler, just one `remove` and then `pop` each dependency, and finally `pop` the `New`.

It might be useful to have a `removeFully` method on `Item` which calls `remove` for that item and then `pop` on each dependency. It would only be used by these two nodes (for now) but the fact that the behavior is both simple and well-defined means it would fit in, and it might be useful for debugging as well.

Ladicek: I'm fairly sure I can reliably find the insertion point for the `Pop` node, at least for the 2 cases I handle here (I'm not aware of any other possible cases that would need similar handling), because it's always after the initial operation and it's created in the same `BlockCreator` method.

But you're right that representing the entire result as an `Item` whose `pop()` would just remove itself and then pop the dependencies is likely a better solution. Not sure I'll be able to do that myself, but I'll try :-)

src/main/java/io/quarkus/gizmo2/MethodTyped.java
Ladicek: I called it `parameterList()` because it delegates to a method called `parameterList()`, like all other methods in this interface. I can rename it, but preserving the name felt best.

src/main/java/io/quarkus/gizmo2/impl/Item.java
Ladicek: Done, created `NewResult` and `NewArrayResult` that represent the result of the composite operation and are at the end of the [part of the] instruction list, so with correct `forEachDependency()` implementation, inserting `Pop` works just like for other item.s

PR #374, opened 2025-05-12
Rework item binding

### LINKED ISSUE #372: array indexing fails when the index is a bound item
### Issue description

This branch contains a reproducer: https://github.com/Ladicek/gizmo/commits/array-indexing-bug/

dmlloyd:
An item is bound if it is added to the list. Constants are never bound. Derefs are bound once they have been added iff the thing they refer to is bound. Bound items may be used only once. Derefs of constants (or derefs of deref of constants etc.) can be reused many times.

Fixes #372

### DIFF
--- a/src/main/java/io/quarkus/gizmo2/Expr.java
+++ b/src/main/java/io/quarkus/gizmo2/Expr.java
@@ -16,7 +16,8 @@ public sealed interface Expr extends SimpleTyped permits Const, Assignable, This
     ClassDesc type();
 
     /**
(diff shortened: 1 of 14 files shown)


Ladicek:
Your code LGTM, but my commit is not directly mergeable :-) Feel free to squash them, or take the commit from https://github.com/Ladicek/gizmo/commits/array-indexing-bug_v2/

Ladicek (approved):

PR #379, opened 2025-05-13
Introduce revoke node operation

dmlloyd:
When a value is being popped, and the creation of the value involves void-typed operations whose only side effect is to mutate or set up that value, then we can revoke the void-typed nodes and pop their dependencies. In some cases the entire operation can be erased, if the mutation operation values do not have any side effects.

Also fixed a bug in `Dup#pop` where the node returned can be null due to excising the node from the list before capturing its `prev` pointer.


Ladicek (approved):
I don't think this makes too big of a difference in practice, but why not :-)

PR #384, opened 2025-05-15
Create flags and access levels abstraction

### LINKED ISSUE #335: avoid leaking the ClassFile API for working with modifiers (`AccessFlag`)
### Issue description

_No response_

dmlloyd:
Fixes #335.

Version 1 of this change used type-safe constants for all of the modifiers and access levels; it was pretty nifty but added about 30-40 new types, so it seemed a bit to heavy to justify.

This version makes a fair effort to put convenience methods in the right place, but otherwise does all checking at run time.

At present, if you try to introduce a flag to a creator which always forces that flag to be set, then you will get an exception. I'm undecided whether this is a preferable behavior.

I did not do anything about default access levels or modifiers yet, other than more or less mirroring what was already in the code.

### DIFF
--- a/src/main/java/io/quarkus/gizmo2/creator/Access.java
+++ b/src/main/java/io/quarkus/gizmo2/creator/Access.java
@@ -1,0 +1,119 @@
+package io.quarkus.gizmo2.creator;
+
+import static io.github.dmlloyd.classfile.ClassFile.*;
+import static io.quarkus.gizmo2.creator.ModifierLocation.*;
+
+import java.util.EnumSet;
+import java.util.List;
+import java.util.Set;
+
+/**
+ * The possible access levels for an item.
+ */
+public enum Access implements Modifier {
+    /**
+     * The {@code public} access level.
+     */
+    PUBLIC(ACC_PUBLIC, EnumSet.of(
+            INTERFACE_CONCRETE_METHOD,
+            INTERFACE_ABSTRACT_METHOD,
+            INTERFACE_STATIC_FIELD,
+            INTERFACE_STATIC_METHOD,
+            CLASS_CONSTRUCTOR,
+            CLASS_CONCRETE_METHOD,
+            CLASS_ABSTRACT_METHOD,
+            CLASS_NATIVE_METHOD,
+            CLASS_STATIC_METHOD,
+            CLASS_INSTANCE_FIELD,
+            CLASS_STATIC_FIELD,
+            CLASS,
+            INTERFACE,
+            NESTED_CLASS,
+            NESTED_INTERFACE)),
+    /**
+     * The {@code protected} access level.
+     */
+    PROTECTED(ACC_PROTECTED, EnumSet.of(
+            CLASS_CONSTRUCTOR,
+            CLASS_CONCRETE_METHOD,
+            CLASS_ABSTRACT_METHOD,
+            CLASS_NATIVE_METHOD,
+            CLASS_STATIC_METHOD,
+            CLASS_INSTANCE_FIELD,
+            CLASS_STATIC_FIELD,
+            NESTED_CLASS,
+            NESTED_INTERFACE)),
+    /**
+     * The package-private ("default") access level.
+     */
+    PACKAGE_PRIVATE(0, EnumSet.of(
+            CLASS_CONSTRUCTOR,
+            CLASS_CONCRETE_METHOD,
+            CLASS_ABSTRACT_METHOD,
+            CLASS_NATIVE_METHOD,
+            CLASS_STATIC_METHOD,
+            CLASS_INSTANCE_FIELD,
+            CLASS_STATIC_FIELD,
+            CLASS,
+            INTERFACE,
+            NESTED_CLASS,
+            NESTED_INTERFACE)),
+    /**
+     * The {@code private} access level.
+     */
+    PRIVATE(ACC_PRIVATE, EnumSet.of(
+            INTERFACE_CONCRETE_METHOD,
+            INTERFACE_STATIC_METHOD,
+            CLASS_CONSTRUCTOR,
+            CLASS_CONCRETE_METHOD,
+            CLASS_NATIVE_METHOD,
+            CLASS_STATIC_METHOD,
+            CLASS_INSTANCE_FIELD,
+            CLASS_STATIC_FIELD,
+            NESTED_CLASS,
+            NESTED_INTERFACE));
+
+    private final int mask;
+    private final Set<ModifierLocation> validLocations;
+
+    /**
+     * The list of access values.
+     */
+    public static final List<Access> values = List.of(values());
+
+    Access(final int mask, final Set<ModifierLocation> validLocations) {
+        this.mask = mask;
+        this.validLocations = validLocations;
+    }
+
+    public boolean validIn(ModifierLocation location) {
+        return validLocations.contains(location);
+    }
+
+    public int mask() {
+        return mask;
+    }
+
+    /**
+     * {@return the access level indicated by the given bit mask}
+     * If multiple access bits are present, then the access level with the highest bit value is returned.
+     *
+     * @param bits the bit mask
+     */
+    public static Access of(int bits) {
+        int hob = Integer.highestOneBit(bits & fullMask());
+        return switch (hob) {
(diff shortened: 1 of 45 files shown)


src/main/java/io/quarkus/gizmo2/creator/ModifiableCreator.java
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    /**
+     * Remove the given modifier flags from this creator.
+     *
+     * @param flags the flags to remove (must not be {@code null})
+     * @throws IllegalArgumentException if this creator does not support one of the given flags
+     */
+    default void withoutFlag(ModifierFlag... flags) {
+        withFlags(Arrays.asList(flags));
```
Ladicek, line 68:
```suggestion
    default void withoutFlags(ModifierFlag... flags) {
        withoutFlags(Arrays.asList(flags));
```

src/main/java/io/quarkus/gizmo2/impl/AbstractMethodCreatorImpl.java
```diff
@@ -1,21 +1,20 @@
...
 import java.util.function.Consumer;
 
 import io.quarkus.gizmo2.creator.AbstractMethodCreator;
+import io.quarkus.gizmo2.creator.ModifierLocation;
 
 public final class AbstractMethodCreatorImpl extends MethodCreatorImpl implements AbstractMethodCreator {
     AbstractMethodCreatorImpl(final TypeCreatorImpl owner, final String name) {
-        super(owner, name, Set.of(ABSTRACT), Set.of(PUBLIC, PROTECTED, SYNCHRONIZED, SYNTHETIC, BRIDGE, ABSTRACT, VARARGS));
+        super(owner, name);
+        flags |= ACC_PUBLIC | ACC_ABSTRACT;
```
Ladicek, line 13: `ACC_PUBLIC` is new here, is that intentional?

src/main/java/io/quarkus/gizmo2/impl/AnonymousClassCreatorImpl.java
```diff
@@ -32,6 +36,7 @@ public final class AnonymousClassCreatorImpl extends ClassCreatorImpl implements
     AnonymousClassCreatorImpl(final ClassDesc type, final ClassOutput output, final ClassBuilder zb,
             final ConstructorDesc superCtor, final ArrayList<Expr> captureExprs) {
         super(type, output, zb);
+        flags |= ACC_FINAL;
```
Ladicek, line 39: I see this was removed from `BlockCreatorImpl.lambda()` and added here, but this also affects `BlockCreator.newAnonymousClass()`, which maybe isn't expected?

src/main/java/io/quarkus/gizmo2/creator/Access.java
```diff
@@ -0,0 +1,119 @@
...
+import static io.quarkus.gizmo2.creator.ModifierLocation.*;
+
+import java.util.EnumSet;
+import java.util.List;
+import java.util.Set;
+
+/**
+ * The possible access levels for an item.
+ */
+public enum Access implements Modifier {
```
Ladicek, line 13: I'd maybe call this `AccessLevel`, but your call.

src/main/java/io/quarkus/gizmo2/creator/ModifierFlag.java
```diff
@@ -0,0 +1,107 @@
...
+
+import static io.github.dmlloyd.classfile.ClassFile.*;
+import static io.quarkus.gizmo2.creator.ModifierLocation.*;
+
+import java.util.EnumSet;
+
+/**
+ * A modifier for a type or member.
+ */
+public enum ModifierFlag implements Modifier {
```
Ladicek, line 11: I'd personally love this to be called just `Modifier`, but I have no idea what current `Modifier` should be called. `AbstractModifier` is not great, but I cannot think of a better name ATM.

Ladicek: Alternatively, looking at both `Access` and `ModifierFlag`, I don't think there's too big of a value in keeping them separate. We could just merge them.

src/main/java/io/quarkus/gizmo2/impl/AbstractMethodCreatorImpl.java
dmlloyd: Yeah, I thought package-private is not a very good default here. So I went a little bit back on my "don't change the defaults" promise I guess. But we should revisit it either way with #222.

src/main/java/io/quarkus/gizmo2/creator/Access.java
dmlloyd: Sure, I can do that.

src/main/java/io/quarkus/gizmo2/impl/AnonymousClassCreatorImpl.java
Ladicek: Good point.

src/main/java/io/quarkus/gizmo2/impl/AbstractMethodCreatorImpl.java
Ladicek: I agree we should revisit holistically, but changing the default just for abstract methods seems strange.

src/main/java/io/quarkus/gizmo2/creator/ModifierFlag.java
Ladicek: Yeah, agree that separating them is useful. But it brings the naming issue; `ModifierFlag` is really not a good name.

src/main/java/io/quarkus/gizmo2/impl/AbstractMethodCreatorImpl.java
dmlloyd: Sure, my thinking was just that we didn't explicitly decide to make this package-private, but rather we forgot to have a default. And so the default was "default" just by chance.

But I can take this out if it's important.

Ladicek: What I find important is we don't forget about this :-) No need to take it out right now.

src/main/java/io/quarkus/gizmo2/creator/ModifierFlag.java
Ladicek: How about we call this `Modifier` and what's now called `Modifier` would be called `Flag`? Then, we would have 2 kinds of flags: modifiers and access levels. The field to store both kinds is already called `flags`.

dmlloyd: Well a "flag" is a binary indicator (this definition goes back decades), whereas both access levels and the boolean modifiers are considered modifiers according to the JVMS, so that seems backwards to me in that regard since an access level is not a flag but it is a modifier.

Ladicek: Technically access levels are also flags then, because there's one bit position for each of them. The JVMS actually uses the term flag for access levels too.

However, if we really need to differentiate access levels and other modifiers, I think I'd be fine with calling this `BooleanModifier`, `BinaryModifier`, `SimpleModifier`, `PropertyModifier` or something like that. If we could avoid using the term "flag" in this context completely, that would be a nice bonus :-)

dmlloyd: Access levels are encoded as flags, but they aren't flags, because the *four* access levels are represented by *three* flags which are mutually exclusive. They're flags in the JVM, but not in the language.

I don't know what's wrong with the word "flag" as it has a clear meaning and historical relevance in computing as well.

Ladicek: The only thing wrong about "flags" is that in the JLS and JVMS and practically everywhere in Java, they are a synonym of "modifiers".

Now, practically speaking, I feel bad about stalling this PR for so long. Since this is the only objection unresolved, I'm gonna cave in and approve this PR so you can merge. Let's get this in and perfect it later on :-)

Ladicek (approved):

PR #386, opened 2025-05-16
Gizmo user manual (WIP)

dmlloyd:
Add a user manual for Gizmo 2.x.

### DIFF
--- a/MANUAL.adoc
+++ b/MANUAL.adoc
@@ -2566,6 +2566,65 @@ public class BreakExample {
 [id=return]
 === `return_`
 
+The `return_(...)` operation immediately returns from the current method or constructor.
+If an argument is given, the <<types,type>> of the argument must match the return type of the method.
+Methods that do not return <<void-expr,`void`>> _must_ provide an argument.
+The `return_(...)` operation cannot be used within initializers.
+
+.Example of `return_`
+[source,java]
+----
+import java.nio.file.Path;
+
+import io.quarkus.gizmo2.ClassOutput;
+import io.quarkus.gizmo2.Gizmo;
+import io.quarkus.gizmo2.LocalVar;
+import io.quarkus.gizmo2.ParamVar;
+import io.quarkus.gizmo2.desc.MethodDesc;
+
+public class ReturnExample {
+    private static final MethodDesc parseLong = MethodDesc.of(
+        Long.class,
+        "parseLong",
+        long.class,
+        String.class
+    );
+
+    public static void main(String[] ignored) {
+        Gizmo gizmo = Gizmo.create(ClassOutput.fileWriter(Path.of("output")));
+        gizmo.class_("Return", cc -> {
+            // compute a factorial
+            MethodDesc factorial = cc.staticMethod("factorial", mc -> {
+                mc.returning(long.class);
+                ParamVar n = mc.parameter("n", long.class);
+                mc.body(b0 -> {
+                    b0.ifElse(b0.eq(n, 0), b1 -> {
+                        b1.return_(1);
+                    }, b1 -> {
+                        // return n * factorial(n - 1)
+                        b1.return_(b1.mul(n, b1.invokeStatic(mc.desc(), b1.sub(n, 1))));
+                    });
+                });
+            });
+            // a nice main method
+            cc.staticMethod("main", mc -> {
+                ParamVar args = mc.parameter("args", String[].class);
+                mc.body(b0 -> {
+                    b0.ifNot(b0.eq(args.length(), 1), b1 -> {
+                        b1.throw_(IllegalArgumentException.class, "Expected one argument");
+                    });
+                    LocalVar n = b0.localVar("n", b0.invokeStatic(parseLong, args.elem(0)));
+                    LocalVar res = b0.localVar("res", b0.invokeStatic(factorial, n));
+                    b0.printf("The factorial of %d is %d%n", n, res);
+                    // void methods return nothing
+                    b0.return_();
+                });
+            });
+        });
+    }
+}
+----
+
 [id=throw]
 === `throw_`
 
--- a/MANUAL.adoc
+++ b/MANUAL.adoc
@@ -15,6 +15,11 @@
 [id=classoutput]
 === `ClassOutput`
 
+Implementations of the `ClassOutput` interface receive the bytes of the generated classes and resources.
+
+Users may implement this interface directly or use a provided implementation.
+
+The `ClassOutput.fileWriter(Path)` static factory method returns a `ClassWriter` which writes classes and resources under the given target path.
 
 [id=descriptors]
 == Descriptors
--- a/MANUAL.adoc
+++ b/MANUAL.adoc
@@ -1355,6 +1355,15 @@ This means that it is always safe to directly return the result of a method invo
 [id=yielding]
 === Yielding values from blocks
 
+Sometimes, a block is treated like an expression, like the conditional part of an `if_`, or a `case_` of a `switch_` expression.
+In these cases, the block must yield a value upon exit.
+This can be done using the `yield(...)` method on `BlockCreator`.
+
+If a block has a non-`void` output type, and control leaves the bottom of the block without a `yield(...)`, an exception will be thrown.
+
+Blocks which are not treated like expressions generally have a type of `void`.
+It is permissible to yield the `void` constant (`Const.ofVoid()`) from these blocks, but it is not necessary to do so.
+
 [id=arithmetic]
 === Arithmetic expressions
 
--- a/MANUAL.adoc
+++ b/MANUAL.adoc
@@ -2515,8 +2515,13 @@ public class AnonymousClassExample {
 [id=control]
 == Control flow
 
+Gizmo provides many control flow constructs comprising a superset of what is available in the Java language.
+
 [id=block]
 === `block`
+
+The `block(...)` operation constructs a new block with its own scope.
+See <<blocks>> for information about creating and using blocks.
 
 [id=break]
 === `break`
--- a/MANUAL.adoc
+++ b/MANUAL.adoc
@@ -50,6 +50,11 @@ public class HelloWorld {
 [id=gizmo-interface]
 == The `Gizmo` interface
 
+The `Gizmo` interface is the primary entry point into the API.
+Instances are created using the `Gizmo.create(...)` method, which accepts a <<classoutput>> instance.
+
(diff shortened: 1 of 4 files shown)


MANUAL.adoc
```diff
@@ -59,18 +59,121 @@ The `ClassOutput.fileWriter(Path)` static factory method returns a `ClassWriter`
...
 == Descriptors
 
+The class file format itself encodes types and members using "descriptors", which are strings which unambiguously identify them.
+Gizmo 2 uses object abstractions of these strings for convenience and safety.
+Some of these object types are provided by the JDK, and others are defined by Gizmo.
+
 [id=classdesc]
 === The `ClassDesc` interface
 
+Types (classes and interfaces, and their variations) are identified using the `ClassDesc` interface, which is provided by the JDK as part of the `java.lang.constant` package.
```
Ladicek, line 69:
```suggestion
Types (primitives, classes and interfaces, and their variations) are identified using the `ClassDesc` interface, which is provided by the JDK as part of the `java.lang.constant` package.
```

Ladicek:
Added a few comments for _Descriptors_, other chapters look fine to me.

MANUAL.adoc
```diff
@@ -0,0 +1,419 @@
...
+
+To get started quickly, include the `io.quarkus.gizmo:gizmo2` artifact in your project build. The following code blob shows how to get started quickly with generating classes:
+
+."Hello world" example
+[source,java]
+----
+public class HelloWorld {
+    public static void main(String[] args) {
+        Gizmo gizmo = Gizmo.create(ClassOutput.fileWriter(Path.of("output")));
+        gizmo.class_("HelloWorldGenerated", zc -> {
```
mkouba, line 28: Why `zc`?

Ladicek: I tend to use `cc` myself, but then when I generate a constructor, I cannot use `cc` again, so I use `mc` even in constructors...

dmlloyd: Yeah I use `cc` for constructors, so then I use `zc` as a mnemonic abbreviation for `clazz creator`, which is a traditional variable name for `Class`-typed things.

Another option is to use `mc` for methods *and* constructors by convention, freeing `cc` for classes.

MANUAL.adoc
```diff
@@ -0,0 +1,419 @@
...
+
+[id=quick-start]
+== Quick start
+
+To get started quickly, include the `io.quarkus.gizmo:gizmo2` artifact in your project build. The following code blob shows how to get started quickly with generating classes:
+
+."Hello world" example
+[source,java]
+----
+public class HelloWorld {
```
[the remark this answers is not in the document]
dmlloyd: You're right; I did start doing that later on.

MANUAL.adoc
```diff
@@ -190,6 +190,87 @@ MethodTypeDesc mainType = MethodTypeDesc.of(
...
+import io.quarkus.gizmo2.ClassOutput;
+import io.quarkus.gizmo2.ClassVersion;
+import io.quarkus.gizmo2.Gizmo;
+import io.quarkus.gizmo2.creator.AccessLevel;
+import io.quarkus.gizmo2.creator.ModifierFlag;
+
+public class SimpleInterfaceExample {
+    public static void main(String[] args) {
+        Gizmo gizmo = Gizmo.create(ClassOutput.fileWriter(Path.of("output")));
+        // create a new class using the given lambda
```
Ladicek, line 254:
```suggestion
        // create a new interface using the given lambda
```

MANUAL.adoc
```diff
@@ -274,6 +274,79 @@ public class SimpleInterfaceExample {
...
 
+Fields may be added to types using the `field`, `staticField`, and `constantField` methods.
+These methods accept a consumer for a `StaticFieldCreator` or `InstanceFieldCreator` instance.
+The `field` method returns a <<fielddesc,`FieldDesc`>> for the newly created field.
+The `staticField` and `constantField` methods return a <<fields,`StaticFieldVar`>> which may be used directly in expressions.
+
+Fields may be initialized directly with a constant value at the time the field is defined by using the `setInitial` method.
+For more complex initializers, the `setInitializer` method may be used, which accepts a <<blocks,code block creator>> which must yield a value for the field.
+Initializers are automatically emitted into constructors or static initializers in the order that they are established.
+
```
Ladicek, line 285: Since the "Creating types" chapter says

> By convention, the parameter name `cc` is typically used for the `ClassCreator` or `InterfaceCreator` instance.

maybe we should say the similar thing here for field creators?

> By convention, the parameter name `fc` is typically used for the `FieldCreator` instance.

dmlloyd:
I've added `Blocks` and `Scoping` today.

MANUAL.adoc
```diff
@@ -806,6 +806,40 @@ public class NaturalOrderingExample {
 [id=types]
 === Types
 
+Every expression has a _type_.
+The type of an expression is generally represented by a <<classdesc,`ClassDesc` descriptor>>.
+The type may be primitive such as `int` or `boolean`, an object type like `java.lang.Object`, or `void`.
```
Ladicek, line 811: Also array types.

MANUAL.adoc
```diff
@@ -806,6 +806,40 @@ public class NaturalOrderingExample {
...
+In many cases, the type of operations can be _inferred_ based on their inputs.
+In some cases, like method invocations or field accesses, the type of the expression is derived from the type of the member.
+In other cases, the type must be specified explicitly.
+In these cases, there will usually be an overloaded variation of the operation which accepts `Class<?>` instead of `ClassDesc` (or a list thereof).
+
+The type of an expression can be queried using the `type()` method of `Expr` and other typed entities (which are all subtypes of `SimpleTyped`).
+
+It is sometimes useful to categorize types as either being primitive or non-primitive.
+This is done using the `typeKind()` method, which is also found on `SimpleTyped` and its subtypes, including `Expr`.
+The return type of this method is `TypeKind`, an enumeration which contains constants for each of the eight primitive types, one for object types, and one for `void`.
```
Ladicek, line 823:
```suggestion
The return type of this method is `TypeKind`, an enumeration which contains constants for each of the eight primitive types, one for reference types, and one for `void`.
```

MANUAL.adoc
```diff
@@ -849,6 +849,27 @@ Most constants can be created once and reused as many times as needed.
 [id=const]
 ==== The `Const` interface
 
+The `Const` interface is used to represent expressions which are constant.
+This interface also contains factory methods to create new constants, including (but not limited to):
+
+* Primitive values (e.g. integers, booleans)
+* Primitive wrapper type instances (e.g. `Long`, `Double`)
```
Ladicek, line 856: I don't think we have that? The factory methods take wrapper types (such as `Boolean`), but always produce primitive constants (such as `boolean`).

MANUAL.adoc
```diff
@@ -849,6 +849,27 @@ Most constants can be created once and reused as many times as needed.
...
+* `enum` constants (as <<jdk-const,`Constable`>> instances)
+* Strings
+* Classes (as `Class<?>` or as `ClassDesc`)
+* `null`
+* The value of a static final field
+* Lists of other object-typed constants (as an immutable `List`)
+* Maps of other object-typed constants (as an immutable `Map`)
+* Other <<jdk-const,JDK constant types>> such as method handles or dynamic constants
+
```
Ladicek, line 865: Maybe also mention `Const.ofDefault()` below this list, as a way to obtain a default value of given type (zero for primitives, `null` for references)?

dmlloyd:
I pushed up fixes to the code examples so that they all have `main` methods and, when possible, also create classes with `main` methods. This makes it a little easier to test.

I've also added the following sections:

* Yielding values from blocks
* Arithmetic expressions
* Bitwise logical expressions
* Bitwise shift expressions
* Equality and ordering operators
* Comparator-like operations
* Invocation expressions

MANUAL.adoc
```diff
@@ -1657,7 +1657,198 @@ public class NegDoubleExample {
...
+        });
+    }
+}
+----
+
+[id=xor]
+==== `xor`
+
+The `xor(...)` operation performs a bitwise-XOR on the two arguments, returning the result.
+The Java language equivalent of this operation is the `|` operator.
```
Ladicek, line 1767:
```suggestion
The Java language equivalent of this operation is the `^` operator.
```

Ladicek:
Additions LGTM, just a small note: I'm not sure about the claims of commutativity of arithmetic operations in case of floating points. It also feels like the comments about order of operands being [in]significant are superfluous; generally, order of operands is always significant. I'd just drop those lines, but leaving the decision up to you.

MANUAL.adoc
```diff
@@ -2693,7 +2693,58 @@ public class GotoExample {
...
+                    b0.ifNot(b0.eq(args.length(), 2), b1 -> {
+                        b1.throw_(IllegalArgumentException.class, "Expected 2 arguments");
+                    });
+                    LocalVar len0 = b0.localVar("len0", b0.withString(args.elem(0)).length());
+                    LocalVar len1 = b0.localVar("len1", b0.withString(args.elem(1)).length());
+                    b0.if_(b0.eq(len0, len1), b1 -> {
+                        b1.printf("Both arguments have the same length%n");
+                    });
+                    b0.ifElse(b0.lt(args.elem(0), args.elem(1)), b1 -> {
+                        b1.printf("Argument 0 comes before argument 1");
```
Ladicek, line 2731:
```suggestion
                        b1.printf("Argument 0 comes before argument 1%n");
```

MANUAL.adoc
```diff
@@ -2693,7 +2693,58 @@ public class GotoExample {
...
+                    }, b1 -> {
+                        b1.printf("Argument 0 does not come before argument 1");
```
Ladicek, line 2733:
```suggestion
                        b1.printf("Argument 0 does not come before argument 1%n");
```

Ladicek:
Additions LGTM. Great examples BTW!

gsmet:
FWIW even if incomplete, I think it would be worth merging to make it more accessible.

dmlloyd:
Fixed the aformentioned problems, plus some bad import formatting, and added:

* Logical and/or
* Switch
* gotoCase/gotoDefault
* loop

I'd ask that any further edits, fixes, and additions be done as a separate PR.
The manual is about 90% done but I cannot work on it further for a few weeks.

Thanks for all the reviews!

gsmet:
> It is said right in the first sentence, but it's small and I think a lot of people will miss it.

We need MOOOORE examples!

PR #393, opened 2025-05-22
improve local vars

Ladicek:
The `define()` method is now called `declare()`, because there's no distinction between declarations and definitions in Java (actually there are no definitions; it's just declarations).

Also, variables that were previously declared but not initialized are now initialized to the default value of their type.

### DIFF
--- a/src/main/java/io/quarkus/gizmo2/creator/BlockCreator.java
+++ b/src/main/java/io/quarkus/gizmo2/creator/BlockCreator.java
@@ -122,8 +122,8 @@ public sealed interface BlockCreator extends SimpleTyped permits BlockCreatorImp
     // lexical variables
 
     /**
-     * Declare a local variable, but doesn't assign it a value.
-     * Such variables may not be read before they are written.
+     * Declare a local variable with given {@code name} and {@code type}, which is initialized
+     * to the {@linkplain Const#ofDefault(ClassDesc) default value} of given type.
      * <p>
      * Variable names are not strictly required to be unique, but it is a good practice.
      *
@@ -134,8 +134,8 @@ public sealed interface BlockCreator extends SimpleTyped permits BlockCreatorImp
     LocalVar declare(String name, GenericType type);
 
     /**
-     * Declare a local variable, but doesn't assign it a value.
-     * Such variables may not be read before they are written.
+     * Declare a local variable with given {@code name} and {@code type}, which is initialized
+     * to the {@linkplain Const#ofDefault(ClassDesc) default value} of given type.
      * <p>
      * Variable names are not strictly required to be unique, but it is a good practice.
      *
@@ -148,8 +148,8 @@ public sealed interface BlockCreator extends SimpleTyped permits BlockCreatorImp
     }
 
     /**
-     * Declare a local variable, but doesn't assign it a value.
-     * Such variables may not be read before they are written.
+     * Declare a local variable with given {@code name} and {@code type}, which is initialized
+     * to the {@linkplain Const#ofDefault(Class) default value} of given type.
      * <p>
      * Variable names are not strictly required to be unique, but it is a good practice.
      *
@@ -162,7 +162,8 @@ public sealed interface BlockCreator extends SimpleTyped permits BlockCreatorImp
     }
 
     /**
-     * Declare a local variable and define its initial value.
+     * Declare a local variable with given {@code name}, which is initialized to the given {@code value}.
+     * The type of the new variable is the type of the {@code value}.
      * <p>
      * Variable names are not strictly required to be unique, but it is a good practice.
      *
@@ -170,23 +171,22 @@ public sealed interface BlockCreator extends SimpleTyped permits BlockCreatorImp
      * @param value the variable initial value (must not be {@code null})
      * @return the local variable (not {@code null})
      */
-    default LocalVar define(String name, Expr value) {
+    default LocalVar declare(String name, Expr value) {
         LocalVar var = declare(name, value.genericType());
         set(var, value);
         return var;
     }
 
     /**
-     * Declare a local variable which is initialized to the given variable's current value.
+     * Declare a local variable, which is initialized to the given variable's current value.
+     * The name and type of the new variable are the same as the name and type of the {@code original}.
      * The given variable may be a parameter, local variable, or field variable.
-     * <p>
-     * The name of the new variable is the same as the name of the {@code original}.
      *
      * @param original the original variable (must not be {@code null})
-     * @return the new local variable (must not be {@code null})
+     * @return the new local variable (not {@code null})
      */
-    default LocalVar define(Var original) {
-        return define(original.name(), original);
+    default LocalVar declare(Var original) {
+        return declare(original.name(), original);
     }
 
     // reading memory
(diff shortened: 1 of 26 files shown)


dmlloyd:
This change introduces an inconsistency: if I `declare("foo", String.class)` I might reasonably expect an uninitialized `String`-type variable called "foo", *or* an initialized `Class`-type variable called "foo" whose initial value is `String.class`.

I do think we need to distinguish between the case of setting up an uninitialized variable versus initializing it immediately (whatever you want to call it).

dmlloyd:
> The initialization of non-initialized local vars is a deviation from standard Java behavior, so I'd be fine with backing that one out; it just felt sensible.

I think this is probably fine actually. For simplicity we scope all variables from the declaration through the end of its containing block; however this means that the LVT might contain entries for variables that don't exist yet (resulting in transient "Incorrect debug information" messages in the debugger) or which may even contain stale values (which is even more confusing).

Any JIT would be intelligent enough to eliminate useless stores, so having the default value be zero is perfectly reasonable as it solves both of the above issues without introducing much practical overhead (other than the zero-init sequence itself, which in the absolute worst case would be like 5-6 bytecodes if I had to guess; most likely it'd be 2 or 3).

dmlloyd:
> > This change introduces an inconsistency: if I `declare("foo", String.class)` I might reasonably expect an uninitialized `String`-type variable called "foo", _or_ an initialized `Class`-type variable called "foo" whose initial value is `String.class`.
> 
> I admit this didn't even occur to me, and I was originally thinking I wouldn't call that a reasonable expectation -- except we do provide `Class`-typed shortcuts to an `Expr` on a few places, so it may actually be expected.
> 
> On the other hand, we don't provide any other typical shortcut for `declare()` (or `define()`, or whatever), so the expectation would still be strange.

We should though, especially for `int` but probably other things too. I got to a point where I was just tired of making method overloads. :-)

> And the javadoc is fairly clear what `declare(String, Class)` means.
> 
> I can imagine though that this might lead to troubles in certain corner cases, so I ultimately agree with
> 
> > I do think we need to distinguish between the case of setting up an uninitialized variable versus initializing it immediately (whatever you want to call it).
> 
> I'm not sure of a good name that includes `declare` at the beginning. Something like `declareAndSet()` or `declareWithInitial()` (the `AndSet` variant is probably preferrable because it's a lot shorter, but it's still a lot longer than `define`)...

How about turning it around? Since initializing the variable immediately seems more common (so far) we could have `declare(name, val)` and `declareZero(name, type)` or something like that.

Ladicek:
I think we could call them `declareUnset()` and the opposite would logically be `declareSet()`. One of them could be called just `declare()`, but I don't really mind which one.

dmlloyd:
> I think we could call them `declareUnset()` and the opposite would logically be `declareSet()`. One of them could be called just `declare()`, but I don't really mind which one.

It's not really unset though, it's set to the default value, so that's a little misleading.

What if we got rid of uninitialized variables altogether? You would always give a value, possibly zero/null.

dmlloyd:
Then perhaps we could call the method `localVar` instead of `declare`. That would make the whole thing a bit clearer IMO.

Ladicek:
That makes perfect sense, actually. +1, I'll do that.

mkouba:
> Then perhaps we could call the method `localVar` instead of `declare`. That would make the whole thing a bit clearer IMO.

+1 for `localVar()`. Maybe even `var()` or `newVar()`?

Ladicek:
Done.

I had to leave an overload that accepts a `GenericType`, because most `Expr`s have a `genericType()` that returns an erasure (`GenericType.of(type())`). If/when we propagate generic types through all `Expr`s (notably, fields), we wouldn't need the explicitly typed method at all -- but for now, we do.

With that, I also felt compelled to leave an overload that accepts a `ClassDesc` and also a `Class` as a type. But I have no hard feelings about these 2 overloads; if you feel we don't need them, I can delete them.

Then, there's an overload that accepts a name and an `Expr`, which takes the type from the `Expr` as expected. (As mentioned above, that type is often an erasure.)

I also added overloads that accept `boolean`, `int`, `long`, `float`, `double` and `String` constants directly. I also tried adding an overload that accepts a `Class` constant, but it felt weird (wasn't sure if that `Class` is a _type_ of the local var, or its _value_), so I got rid of that.

mkouba:
Hm, sorry for my ignorance but why did we remove the possibility to declare a local variable but don't assign it a value? It might be useful, or am I missing something?

Ladicek:
Fortunately in Gizmo, we can write that in a more direct fashion. There's no corresponding Java syntax, but it would look roughly like:

```java
final int val = {
    if (something) {
        yield 1;
    } else {
        yield 2;
    }
};
```

Ladicek:
Actually, can we declare `final` local vars in Gizmo? I don't think so...

mkouba:
> Fortunately in Gizmo, we can write that in a more direct fashion. There's no corresponding Java syntax, but it would look roughly like:
> 
> ```java
> final int val = {
>     if (something) {
>         yield 1;
>     } else {
>         yield 2;
>     }
> };
> ```

It could be more direct but also more cryptic :shrug:

dmlloyd (commented):
> Done.
> 
> I had to leave an overload that accepts a `GenericType`, because most `Expr`s have a `genericType()` that returns an erasure (`GenericType.of(type())`). If/when we propagate generic types through all `Expr`s (notably, fields), we wouldn't need the explicitly typed method at all -- but for now, we do.

It's probably a good idea anyway; sometimes the generic type of a variable differs from the type of its value.

> With that, I also felt compelled to leave an overload that accepts a `ClassDesc` and also a `Class` as a type. But I have no hard feelings about these 2 overloads; if you feel we don't need them, I can delete them.

I think this is fine actually, for the above reason.

> I also added overloads that accept `boolean`, `int`, `long`, `float`, `double` and `String` constants directly. I also tried adding an overload that accepts a `Class` constant, but it felt weird (wasn't sure if that `Class` is a _type_ of the local var, or its _value_), so I got rid of that.

That's fine I think. It's less common than the other types I suspect.

>> Hm, sorry for my ignorance but why did we remove the possibility to declare a local variable but don't assign it a value? It might be useful, or am I missing something?
>
> Yeah, it's useful to create `.class` files that don't pass verification :-) I honestly don't see a reason why we should allow this.

I don't think it creates a verification issue per se, but it does create other problems relating to debugging. The alternative would be to come up with a much more sophisticated local var mapping system (more like what `javac` would do) but I don't think it's worth it in order to support this pattern. As I said above, just initializing to zero is very simple in (I think) all cases, so that solves all the problems at once.

> Actually, can we declare `final` local vars in Gizmo? I don't think so...

There's no way to reflect `final` for local vars in a class file, so that has been left out.  We could conceivably support it though (to an extent), and local variable annotations as well (sort of), by introducing the optional use of a `LocalVarCreator` which handles all that stuff. But this would be very flimsy and a real stretch of the interpretation of the JVMS.

> Fortunately in Gizmo, we can write that in a more direct fashion. There's no corresponding Java syntax, but it would look roughly like: ...

I think for this case it's OK to use a zero-initialized (non-"final") local var. The JVM should be more than capable of dropping the redundant store at even the lowest JIT levels I would expect. And like I said before, if this does become a real problem, we could possibly enhance this with a much fancier first-use-site allocator that allows us to re-add uninitialized local vars, but there's a degree of control flow analysis involved in doing this correctly (and without blowing up the LVT/LVTT attributes with lots of redundant rows), so it doesn't seem worth it for 2.0.

PR #395, opened 2025-05-22
Rename some `with` to `add` or `set`

### LINKED ISSUE #389: A mix of naming strategy
### Issue description

We have a mix of naming strategies for mutator methods, due to the original intent of extending the classfile API rather than hiding it (and thus more or less imitating its idioms).

In some cases, `thing.withXxx` will modify `thing` by adding or changing one of its properties. In other cases, `thing.withXxx` returns a new `Thing` with the added or changed property.

I propose that we adopt this mutator naming strategy:

* Methods which change something should be called `setSomething(newSomething)` (like `fc.setAccess(PUBLIC)`
* Methods which add something to an existing object should be called `addXxxx(xc -> { ... })` (like `zc.addField(fc -> { ... })`
* Methods which remove something from an existing object should be called `removeSomething(something)` (like `fc.removeModifier(TRANSIENT)`)
* Methods which return a new thing with the transformed property should be `Thing newThing = thing.with(newProperty);` (like `Gizmo#withOutput` and probably the solution for #222 as well)

WDYT?

dmlloyd:
Fixes #389

### DIFF
--- a/src/main/java/io/quarkus/gizmo2/GenericType.java
+++ b/src/main/java/io/quarkus/gizmo2/GenericType.java
@@ -371,7 +371,7 @@ public abstract class GenericType {
      */
     public <A extends java.lang.annotation.Annotation> GenericType withAnnotation(Class<A> annotationType,
             Consumer<AnnotationCreator<A>> builder) {
-        return withAnnotations(ac -> ac.withAnnotation(annotationType, builder));
+        return withAnnotations(ac -> ac.addAnnotation(annotationType, builder));
     }
 
     /**
--- a/src/main/java/io/quarkus/gizmo2/TypeArgument.java
+++ b/src/main/java/io/quarkus/gizmo2/TypeArgument.java
@@ -226,7 +226,7 @@ public abstract class TypeArgument {
          * @param annotationType the annotation type
          */
         public <A extends java.lang.annotation.Annotation> OfAnnotated withAnnotation(final Class<A> annotationType) {
-            return withAnnotations(ac -> ac.withAnnotation(annotationType));
+            return withAnnotations(ac -> ac.addAnnotation(annotationType));
         }
 
         /**
@@ -237,7 +237,7 @@ public abstract class TypeArgument {
          */
         public <A extends java.lang.annotation.Annotation> OfAnnotated withAnnotation(final Class<A> annotationType,
                 final Consumer<AnnotationCreator<A>> builder) {
-            return withAnnotations(ac -> ac.withAnnotation(annotationType, builder));
+            return withAnnotations(ac -> ac.addAnnotation(annotationType, builder));
         }
 
         public StringBuilder toString(StringBuilder b) {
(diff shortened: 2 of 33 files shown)


Ladicek:
Here's a summary of what this PR does:

```
AnnotatableCreator.withAnnotation -> addAnnotation
AnnotationCreator.with -> add
AnnotationCreator.withArray -> addArray
ExecutableCreator.withType -> setType
FieldCreator.withType -> setType
FieldCreator.withInitial -> setInitial
FieldCreator.withInitializer -> setInitializer
ModifiableCreator.withFlag -> addFlag
ModifiableCreator.withFlags -> addFlags
ModifiableCreator.withoutFlag -> removeFlag
ModifiableCreator.withoutFlags -> removeFlags
ModifiableCreator.withAccess -> setAccess
ParamCreator.withType -> setType
TypeVariableCreator.withFirstBound -> setFirstBound
TypeVariableCreator.withOtherBounds -> setOtherBounds
```

I'm missing `TypeCreator.withVersion`, but other than that, it seems comprehensive. Do you want to fix that in this PR? It can wait for another, for sure.

A few remarks, but I don't want them to be addressed in this PR: I would personally name the `setType` method (on `ExecutableCreator`, `FieldCreator` and `ParamCreator`) just `type` (together with `setInitial` and `setInitializer` on `FieldCreator`), but that's perhaps just me. Also, I would collapse `setFirstBound` and `setOtherBounds` on `TypeVariableCreator` to `setBounds` (or just `bounds`), but I would like to do more changes on generic types, so let's just ignore that :-)

dmlloyd:
> Here's a summary of what this PR does:
> 
> 
> 
> ```
> 
> AnnotatableCreator.withAnnotation -> addAnnotation
> 
> AnnotationCreator.with -> add
> 
> AnnotationCreator.withArray -> addArray
> 
> ExecutableCreator.withType -> setType
> 
> FieldCreator.withType -> setType
> 
> FieldCreator.withInitial -> setInitial
> 
> FieldCreator.withInitializer -> setInitializer
> 
> ModifiableCreator.withFlag -> addFlag
> 
> ModifiableCreator.withFlags -> addFlags
> 
> ModifiableCreator.withoutFlag -> removeFlag
> 
> ModifiableCreator.withoutFlags -> removeFlags
> 
> ModifiableCreator.withAccess -> setAccess
> 
> ParamCreator.withType -> setType
> 
> TypeVariableCreator.withFirstBound -> setFirstBound
> 
> TypeVariableCreator.withOtherBounds -> setOtherBounds
> 
> ```
> 
> 
> 
> I'm missing `TypeCreator.withVersion`, but other than that, it seems comprehensive. Do you want to fix that in this PR? It can wait for another, for sure.

Sure, I can do that when I get started today.

> A few remarks, but I don't want them to be addressed in this PR: I would personally name the `setType` method (on `ExecutableCreator`, `FieldCreator` and `ParamCreator`) just `type` (together with `setInitial` and `setInitializer` on `FieldCreator`), but that's perhaps just me.

I'll look into that as well this morning.

> Also, I would collapse `setFirstBound` and `setOtherBounds` on `TypeVariableCreator` to `setBounds` (or just `bounds`), but I would like to do more changes on generic types, so let's just ignore that :-)

Unfortunately that would be incorrect. The first bound is special (and optional) not just due to the class file API but also due to requirements of the class file format itself, and also the JLS. Generally the first bound can be a type variable or a class, and the extra bounds must be interfaces (but only if the first bind is not a type variable).

dmlloyd:
The `withVersion` issue is fixed.

PR #402, opened 2025-05-28
Add a mechanism to define the default flags of members

### LINKED ISSUE #222: Gizmo2: consider the default flags for classes and members

dmlloyd:
Add a mechanism to define the default flags of members. Fixes #222.

This is a draft to see if we like this approach.


dmlloyd:
I think the general API is simple enough, and having it be internally uniform keeps things consistent and makes it easier to add new flags in the future (value types? who knows).

The problem with specific methods for categories is the same as the categories themselves: it's a little tricky to get them right, and useful categories might overlap: for example once we introduce nested classes, if I set "classes" to "package-private" I might be surprised that my nested classes didn't get affected - or I might be surprised if they *did*.

By spelling out each thing specifically, we'll never have to deal with these kinds of opinions. And, we keep things internally simple too, since adding creators for new kinds of things would require new modifier locations but otherwise modifiers would just work.

Ladicek (approved):

PR #404, opened 2025-05-28
rethink generic types

Ladicek:
This is a radical proposal to move back to non-generic API using just a single representation of types: `ClassDesc`. The introduction of `GenericType` had a nice effect of generating signatures automatically, but it introduced more complexity (two representations of types instead of one), irregularity (you had to use `GenericType` sometimes and `ClassDesc` other times), while not fully reaping all benefits that a generic type representation would allow when introduced holistically. However, to do that, we would have to use `GenericType` _everywhere_, including class/method/field descriptors, and forget about using `ClassDesc` in the API. That didn't happen, so you could for example introduce a field of type `List<String>`, but when reading from it, you'd get a `List`.

I admit there's likely little to no desire to merge this, but I just had to propose it, because the more I work with the new API, the less I like it.

There are 2 commits:

### remove GenericType and related classes

This commit moves back to a single, non-generic representation of types
using `ClassDesc`. This is faithful to bytecode, which only retains
generics in the `Signature` attribute, but is otherwise non-generic.

The introduction of `GenericType` increased complexity (two representations
of types instead of one) and irregularity (you sometimes have to use
`GenericType`, and other times, you have to use `ClassDesc`), while only
providing mediocre benefits (signatures are generated automatically and
type annotations are possible; on the other hand, you can create a field
of type `List<String>`, but when you read from it, you get a `List`).
To really gain the benefits, we would have to represent types using
`GenericType` _everywhere_, including field/method/type descriptors.
That would be a significant undertaking and the API would be harder
to use (especially with current `GenericType`, which is relatively hard
to construct).

### add specialized API to set class/method/field signatures

The API includes:

- representation of generic types (`GenericType` and all its member types)
- signature creators (`ClassSignatureCreator`, `InterfaceSignatureCreator`,
  `MethodSignatureCreator`, `FieldSignatureCreator`)

Using the API leads to certain amount of information duplicity. This is
unfortunate, but it's the price for not spreading the generic type API
across all of the code base.

This API doesn't include type annotations, which should be relatively
straightforward to add if there's a desire.


dmlloyd:
> That didn't happen, so you could for example introduce a field of type `List<String>`, but when reading from it, you'd get a `List`.

That isn't an inherent design problem, just a small missing API feature. I don't think it would be particularly difficult to propagate the generic types API throughout the rest of the API without introducing any major problems.

Are there other problems that you have with the API?

PR #405, opened 2025-05-29
Allow generic type to be specified in more places

dmlloyd:
Also fix some missed propagations.


Ladicek (approved):

Ladicek:
Actually one thing: we should probably validate that the erasure of the given generic type is equal to the `ClassDesc` of the field descriptor / method return type / etc.

PR #406, opened 2025-05-29
Fix memory order inconsistency with array store

dmlloyd:


dmlloyd:
Also add tests.

Ladicek (approved):
I noticed this, but wasn't sure if that's a limitation of arrays or just an omission. I didn't follow through on that, though. Thanks!

PR #437, opened 2025-07-03
improvements

Ladicek:
- add `BlockCreator.newArray()`, `listOf()` and `setOf()` overloads that take a mapping function
- tiny javadoc improvements
- improve lambda method names in debug mode
- add reference widening conversion, but only to `java.lang.Object`
- add reference narrowing conversion, but only from `java.lang.Object`
- fix `Gizmo.withOutput()`

### DIFF
--- a/src/main/java/io/quarkus/gizmo2/impl/GizmoImpl.java
+++ b/src/main/java/io/quarkus/gizmo2/impl/GizmoImpl.java
@@ -68,7 +68,7 @@ public final class GizmoImpl implements Gizmo {
     }
 
     public Gizmo withOutput(final ClassOutput outputHandler) {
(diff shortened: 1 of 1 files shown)


Ladicek:
I'd like to highlight that for the reference narrowing conversion, I tried to make it opt-in (by adding `Gizmo.withObjectDowncasts()`) and it mostly worked, except I couldn't pass the `Gizmo` object to `ArrayDeref` and `ArrayStore`. Those would require all `Item`s to have an instance of `Gizmo`, which is not possible for `Const`s. Hence, it is enabled always, which is not terribly bad, because Gizmo 1 allowed it too.

dmlloyd (approved):

PR #444, opened 2025-07-16
fix GenericType.toString()

Ladicek:
The following kinds of `GenericType` had incorrect `toString()`:

- primitive types: instead of the Java keyword, the string representation was just a single-character descriptor
- class types: type annotations were on the package name (`@MyAnn com.example.FooBar`), instead of the class name (`com.example.@MyAnn FooBar`)
- array types: type annotations were not on the correct array components

This commit fixes these issues and also makes the `TypeArgument.OfWildcard.copy()` method `abstract`, just because returning `null` makes no sense and all subclasses override the method anyway.


src/main/java/io/quarkus/gizmo2/GenericType.java
```diff
@@ -669,8 +669,25 @@ public ClassDesc desc() {
...
+            appendComponent(b);
+            return b;
+        }
+
+        private void appendComponent(StringBuilder b) {
+            // cannot use just `hasAnnotations()`, because that delegates to `has{Visible,Invisible}Annotations()`,
+            // which are both overridden in this class to also check the component type
+            if (super.hasVisibleAnnotations() || super.hasInvisibleAnnotations()) {
+                b.append(' ');
+            }
```
Ladicek, line 686: This _possibly maybe_ suggests that the overrides of `hasVisibleAnnotations()` and `hasInvisibleAnnotations()` in this class are not correct. They look like:

```java
public boolean hasVisibleAnnotations() {
    return super.hasVisibleAnnotations() || componentType.hasVisibleAnnotations();
}

public boolean hasInvisibleAnnotations() {
    return super.hasInvisibleAnnotations() || componentType.hasInvisibleAnnotations();
}
```

I have no idea why they are like this -- it is my opinion that these methods shouldn't be overridden in this class at all, but I also guess there's a reason behind it.

Ladicek: Sure.

PR #445, opened 2025-07-16
Jandex integration

### LINKED ISSUE #301: Jandex integration
We use Jandex as the language model all across Quarkus, and Gizmo is very often used in Quarkus. We should provide some level of integration, for example:

- create a method from Jandex `MethodInfo` (meaning: use the name, signature, modifiers)
- create a field from Jandex `FieldInfo` (meaning: use the name, type, modifiers)
- add a Jandex `AnnotationInstance` as an annotation
- maybe a few more

This doesn't necessarily have to be in the main Gizmo module, if we don't want a Jandex dependency. We could create a second module just for this.

Ladicek:
Fixes #301


Ladicek:
I didn't even think of putting the Gizmo 2 integration into Jandex, but we can do that, sure. No problem with that.

Ladicek:
Well, OK then. I'll move this to Jandex (not to the core, I'll make a new module there).

PR #449, opened 2025-07-22
Do not validate for a path upfront

dmlloyd:
It defeats the purpose of using `createDirectories` below.


Ladicek (approved):
I don't necessarily think the check is wrong -- it validates the existence of the _base_ directory. If I call `ClassOutput.fileWriter(Paths.get(".../target"))`, it just validates that the `.../target` directory exists. The `Files.createDirectories()` call below makes sure that the directory structure _under_ `.../target` exists.

But then, the check is quite superfluous. We might as well let `Files.createDirectories()` also create `.../target` if it doesn't exist. It doesn't harm anything.

PR #471, opened 2025-08-07
fix `StringBuilderOps`

Ladicek:
The `StringBuilderOps` class didn't guarantee that the object it operates upon is stored in a local variable. This was previously partially fixed by all the `public` methods returning a new instance, but that requires usages to remember that and always reassign. This is error-prone and inconvenient.

With this commit, all constructors of `StringBuilderOps` now guarantee that the object is indeed stored in a local variable, which makes `return this` a valid usage pattern in all the `public` methods. Usage is then easy and less error-prone.

### DIFF
--- a/src/main/java/io/quarkus/gizmo2/creator/ops/StringBuilderOps.java
+++ b/src/main/java/io/quarkus/gizmo2/creator/ops/StringBuilderOps.java
@@ -2,6 +2,7 @@ package io.quarkus.gizmo2.creator.ops;
 
 import io.quarkus.gizmo2.Const;
 import io.quarkus.gizmo2.Expr;
+import io.quarkus.gizmo2.Var;
 import io.quarkus.gizmo2.creator.BlockCreator;
 
 /**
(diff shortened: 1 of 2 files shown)


dmlloyd:
Again reusing *any* `Ops` is not allowed, generally speaking. This class is not different from any other, except that its target methods return the builder instance again as a new expression. That's why returning a new `Ops` instance is correct and would be correct for any such case. There's nothing inherently different about `StringBuilderOps`, nor should there be.

If the restrictions on the `Ops` classes are too onerous, and neither of the two options that I gave above are acceptable, then I'd say we should delete all the `Ops` classes completely and use a different approach, such as keeping constants for frequently-used method descriptors.

Ladicek:
> reusing _any_ `Ops` is not allowed

So first, I think we should document this. It probably feels natural for us by now, but for Gizmo newcomers, it would be an easy mistake.

And second, I'd say, what's wrong with `StringBuilderOps` to say:

> Unlike other `*Ops` classes, instances of `StringBuilderOps` can be reused to append to the same `StringBuilder` in the same `Block` multiple times. This allows using `StringBuilderOps` in the same manner a `StringBuilder` would normally be used.

?

I just don't see the issue. Yes, it is different. Yes, there's a good reason for it to be different.

PR #473, opened 2025-08-08
add automated testing of documentation examples

Ladicek:


gsmet:
This is extremely cool!

Ladicek:
The possibility occurred to me when David started (in later chapters) generating classes with the classic `main` method that reads the provided arguments and writes to stdout. It felt like there's gotta be a way to automatically compile and run the examples and then run the generated classes :-)

gsmet:
Yeah, in Hibernate Validator, we use proper classes for examples and we include the code in the doc.

It's nice because you can have more test infrastructure around. But it's not ideal as you don't have the source handy when you edit the doc files.

The approach you took is perfect for Gizmo's case.

PR #475, opened 2025-08-11
avoid constant dynamic for primitive class constants

Ladicek:
The `Class`-typed constants used to be emitted as LDC always. This leads to constant dynamic being used for primitive types, which we don't want. This commit moves to the classic javac translation strategy: for primitive class constants, we just load the `TYPE` field from the wrapper class.


Ladicek:
For the record, @mkouba had an issue with the original (condy-based) translation strategy with native image. The javac translation strategy apparently works fine.

PR #485, opened 2025-08-20
Avoid instantiating an IllegalStateException in normal case

gsmet:
Except if you're doing something wrong, it is useless and I could actually see it in some profiling.
Using a boolean and only instantiating the IllegalStateException if necessary.

Screenshot From 2025-08-20 16-35-13


gsmet:
Closing as I will group it with some other changes.

dmlloyd:
FWIW it's reasonable. The only reason it was done this way was so that the stack trace will refer to the point where the block was initially completed, which was useful during early debugging, but this is fine too.

PR #488, opened 2025-08-25
Assorted performance improvements

gsmet:
These results are based on the benchmark assembled by @Ladicek. I extract this to some sort of JMH benchmarks.

Baseline with Gizmo 2 (`0dff4ee`) :

```
Result "io.quarkus.arc.crazybeans.test.ArcBuildTimeMeasurementJmhBenchmark.benchmarkArCInitialization":
  0.218 ±(99.9%) 0.003 s/op [Average]
  (min, avg, max) = (0.216, 0.218, 0.223), stdev = 0.002
  CI (99.9%): [0.215, 0.221] (assumes normal distribution)
```

After all these patches:

```
Result "io.quarkus.arc.crazybeans.test.ArcBuildTimeMeasurementJmhBenchmark.benchmarkArCInitialization":
  0.200 ±(99.9%) 0.005 s/op [Average]
  (min, avg, max) = (0.196, 0.200, 0.207), stdev = 0.003
  CI (99.9%): [0.195, 0.205] (assumes normal distribution)
```

Note that we are still far from the results with Gizmo 1 but I need some of my Quarkus patches in to be able to compare things properly between Gizmo 1 and 2.

This should be reviewed commit per commit (and maybe discussed commit per commit).

In some cases, the result is to reduce allocations, in some cases CPU cycles but it all goes in the right direction.

The last one is the most controversial IMHO and I added the commit more to show the issue than to solve the problem this way: we are creating a gazillions of `TypeKind`, in large part because of the calls to `isVoid()`, that then calls `typeKind()` which in most cases instantiates a new `TypeKind`. I think we should cache the value once and for all. In this commit, I fixed the most prominent culprits for ArC but really, I think we should probably change the pattern and enforce the caching for all ``Item``s by having a proper constructor that takes the type as a parameter and caches the type kind value in `Item`.

Happy to discuss these changes in a call if it helps.

### DIFF
--- a/src/main/java/io/quarkus/gizmo2/impl/ConstructorDescImpl.java
+++ b/src/main/java/io/quarkus/gizmo2/impl/ConstructorDescImpl.java
@@ -18,7 +18,7 @@ public final class ConstructorDescImpl implements ConstructorDesc {
         }
         this.owner = owner;
         this.type = type;
-        hashCode = Objects.hash(owner, "<init>", type);
+        hashCode = buildHashCode();
     }
 
     public ClassDesc owner() {
@@ -56,5 +56,12 @@ public final class ConstructorDescImpl implements ConstructorDesc {
 
     public String toString() {
         return toString(new StringBuilder()).toString();
+    }
+
+    public int buildHashCode() {
+        int result = Objects.hashCode(owner);
+        result = 31 * result + "<init>".hashCode();
+        result = 31 * result + Objects.hashCode(type);
+        return result;
     }
 }
--- a/src/main/java/io/quarkus/gizmo2/impl/ExecutableCreatorImpl.java
+++ b/src/main/java/io/quarkus/gizmo2/impl/ExecutableCreatorImpl.java
@@ -30,6 +30,7 @@ import io.github.dmlloyd.classfile.attribute.RuntimeInvisibleTypeAnnotationsAttr
 import io.github.dmlloyd.classfile.attribute.RuntimeVisibleParameterAnnotationsAttribute;
 import io.github.dmlloyd.classfile.attribute.RuntimeVisibleTypeAnnotationsAttribute;
 import io.github.dmlloyd.classfile.attribute.SignatureAttribute;
+import io.github.dmlloyd.classfile.constantpool.ClassEntry;
 import io.quarkus.gizmo2.GenericType;
 import io.quarkus.gizmo2.ParamVar;
 import io.quarkus.gizmo2.This;
@@ -169,8 +170,11 @@ public sealed abstract class ExecutableCreatorImpl extends ModifiableCreatorImpl
         addInvisible(mb);
         List<GenericType.OfThrows> throws_ = this.throws_;
         if (!throws_.isEmpty()) {
-            mb.with(ExceptionsAttribute.of(
-                    throws_.stream().map(GenericType::desc).map(cd -> typeCreator.zb.constantPool().classEntry(cd)).toList()));
+            List<ClassEntry> exceptions = new ArrayList<>(throws_.size());
+            for (int i = 0; i < throws_.size(); i++) {
+                exceptions.add(typeCreator.zb.constantPool().classEntry(throws_.get(i).desc()));
+            }
+            mb.with(ExceptionsAttribute.of(exceptions));
             for (int i = 0; i < throws_.size(); i++) {
                 final GenericType.OfThrows genericType = throws_.get(i);
                 Util.computeAnnotations(genericType, RetentionPolicy.RUNTIME, TypeAnnotation.TargetInfo.ofThrows(i),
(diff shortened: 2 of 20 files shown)


src/main/java/io/quarkus/gizmo2/impl/ExecutableCreatorImpl.java
```diff
@@ -127,8 +128,13 @@ public void setType(final MethodTypeDesc desc) {
 
     MethodTypeDesc computeType() {
         assert type == null;
-        return MethodTypeDesc.of(returnType(),
-                params.stream().map(ParamVarImpl::type).toArray(ClassDesc[]::new));
+
+        ClassDesc[] paramTypes = new ClassDesc[params.size()];
+        for (int i = 0; i < params.size(); i++) {
+            paramTypes[i] = params.get(i).type();
+        }
```
dmlloyd, line 135: I'd be quite surprised to see this actually hit a profile, once the initial lambdas are resolved. How was this benchmarked?

src/main/java/io/quarkus/gizmo2/impl/ConstructorDescImpl.java
```diff
@@ -18,7 +18,7 @@ public ConstructorDescImpl(final ClassDesc owner, final MethodTypeDesc type) {
         }
         this.owner = owner;
         this.type = type;
-        hashCode = Objects.hash(owner, "<init>", type);
+        hashCode = buildHashCode();
```
dmlloyd, line 21: This one I don't quite understand; we were already caching the hash code, so what is the difference here?

src/main/java/io/quarkus/gizmo2/impl/MethodDescImpl.java
```diff
@@ -16,7 +16,7 @@ public sealed abstract class MethodDescImpl implements MethodDesc permits ClassM
         this.owner = owner;
         this.name = name;
         this.type = type;
-        hashCode = Objects.hash(owner, name, type);
+        this.hashCode = buildHashCode();
```
dmlloyd, line 19: Is this array actually visible on profiles as significant?

src/main/java/io/quarkus/gizmo2/impl/ExecutableCreatorImpl.java
gsmet: And AFAICS, I can still see it in AsyncProfiler with the `GenericType#of()` constructor:

Screenshot From 2025-08-25 18-09-46

I'll try dropping it and see if it makes a difference in my JMH benchmark.

gsmet: I can see the method already present and called in Java 21. I'll try to do the benchmark with Java 21 and see if this method is still there.

dmlloyd:
Going through the JDK source, it looks like the descriptor string is cached for both primitive and reference class descriptors, so these should all be inlinable/fast call paths. I wonder if something else might be happening here?

gsmet:
> Going through the JDK source, it looks like the descriptor string is cached for both primitive and reference class descriptors, so these should all be inlinable/fast call paths. I wonder if something else might be happening here?

Yes, only `Class` doesn't have the descriptor string cached and I checked and it's definitely not the culprit.

FWIW, here is the distribution of calls to `SimpledType#typeKind()` in the benchmark. In the benchmark, which generate quite a lot of beans - it has been prepared by Ladislav, we run the generation 30 times. So to get these metrics for one run of the benchmark, you need to split the numbers by 30.

I dropped the numbers that were not really significant.

```
  30748 SimpleTyped#typeKind(): class io.quarkus.gizmo2.impl.constant.StringConst
  30931 SimpleTyped#typeKind(): class io.quarkus.gizmo2.impl.IfRel
  58365 SimpleTyped#typeKind(): class io.quarkus.gizmo2.impl.BlockCreatorImpl
  60000 SimpleTyped#typeKind(): class io.quarkus.gizmo2.impl.CheckCast
  61403 SimpleTyped#typeKind(): class io.quarkus.gizmo2.impl.InstanceOf
  92247 SimpleTyped#typeKind(): class io.quarkus.gizmo2.impl.ThisExpr
 116690 SimpleTyped#typeKind(): class io.quarkus.gizmo2.impl.constant.VoidConst
 193593 SimpleTyped#typeKind(): class io.quarkus.gizmo2.impl.FieldSet
 226562 SimpleTyped#typeKind(): class io.quarkus.gizmo2.impl.FieldDeref
 264637 SimpleTyped#typeKind(): class io.quarkus.gizmo2.impl.IfZero
 419773 SimpleTyped#typeKind(): class io.quarkus.gizmo2.impl.Dup
 781259 SimpleTyped#typeKind(): class io.quarkus.gizmo2.impl.ParamVarImpl
 915867 SimpleTyped#typeKind(): class io.quarkus.gizmo2.impl.Invoke
1329326 SimpleTyped#typeKind(): class io.quarkus.gizmo2.impl.LocalVarSet
1794557 SimpleTyped#typeKind(): class io.quarkus.gizmo2.impl.LocalVarAllocator
3200295 SimpleTyped#typeKind(): class io.quarkus.gizmo2.impl.LocalVarImpl
```

And to `TypeKind#from()`:

```
  11750 Descriptor: ClassOrInterfaceDesc[ContextInstanceHandle]
  15500 Descriptor: ClassOrInterfaceDesc[StartupEvent]
  15501 Descriptor: ClassOrInterfaceDesc[EventContext]
  16134 Descriptor: ClassOrInterfaceDesc[List]
  30656 Descriptor: ClassOrInterfaceDesc[CreationException]
  31000 Descriptor: ClassOrInterfaceDesc[ArcContainer]
  31000 Descriptor: ClassOrInterfaceDesc[InjectableInterceptor]
  46501 Descriptor: ClassOrInterfaceDesc[Supplier]
  46810 Descriptor: ClassOrInterfaceDesc[MyInterceptorBinding_ArcAnnotationLiteral]
  62000 Descriptor: ClassOrInterfaceDesc[ArcUndeclaredThrowableException]
  62000 Descriptor: ClassOrInterfaceDesc[InterceptedMethodMetadata]
  69211 Descriptor: ClassOrInterfaceDesc[String]
  92753 Descriptor: ClassOrInterfaceDesc[ClassLoader]
 114285 Descriptor: ClassOrInterfaceDesc[CreationalContextImpl]
 125014 Descriptor: ClassOrInterfaceDesc[Map]
 139509 Descriptor: ClassOrInterfaceDesc[InjectableContext]
 157051 Descriptor: ClassOrInterfaceDesc[HashMap]
 174597 Descriptor: PrimitiveClassDesc[boolean]
 184177 Descriptor: ClassOrInterfaceDesc[CreationalContext]
 195000 Descriptor: ClassOrInterfaceDesc[MyDependency]
 233066 Descriptor: ClassOrInterfaceDesc[Class]
 247371 Descriptor: ArrayClassDesc[Object, 1]
 248000 Descriptor: ClassOrInterfaceDesc[InjectableBean]
 277962 Descriptor: ClassOrInterfaceDesc[RuntimeException]
 315505 Descriptor: ClassOrInterfaceDesc[Type]
 339274 Descriptor: ClassOrInterfaceDesc[Exception]
1267505 Descriptor: ClassOrInterfaceDesc[Object]
4282593 Descriptor: PrimitiveClassDesc[void]
```

Note that with my patch, we end up having more calls for `ClassOrInterfaceDesc[Object]` probably because we resolve it even in cases where we wouldn't have resolved it by being lazy.

dmlloyd:
Most of the methods involved in `TypeKind.from` are inlinable and the rest are trivial. So I'm not clear on why this shows up as a performance problem. If you remove the call count code, can you show a significant (greater than margin-of-error) difference in performance by caching these? I'm concerned of some larger, hidden issue here.

gsmet:
I will close this one for now as the plan is to extract individual patches if we see an improvement.

PR #493, opened 2025-08-26
Use `String` as map keys instead of `ClassDesc` for generic types and boxing computations

dmlloyd:
This helps to avoid the surprisingly expensive `PrimitiveClassDescImpl.hashCode` and `PrimitiveClassDescImpl.equals` calls (which are expensive because they inherit from `DynamicConstantDesc` which does an excessive amount of work).

This should help make `GenericType.of()` considerably more efficient for the primitive type case.


gsmet:
It seems to improve things a tiny bit but not that much, I will have to dig more about the result of this change.

PR #494, opened 2025-08-26
Only snapshot the path list if needed

dmlloyd:
This should help reduce allocations when emitting `GenericType`.


dmlloyd:
Since allocation is basically paid later at GC, it would only be measurable as slightly longer periods before GC cycles.

PR #496, opened 2025-08-26
Avoid a lot of stream operations in ExecutableCreatorImpl

gsmet:
Note that the point is not only to avoid streams but also to avoid going through the same list again and again.

main

Result "io.quarkus.arc.crazybeans.test.ArcBuildTimeMeasurementJmhBenchmark.benchmarkArCInitialization":
  0.220 ±(99.9%) 0.003 s/op [Average]
  (min, avg, max) = (0.217, 0.220, 0.224), stdev = 0.002
  CI (99.9%): [0.217, 0.224] (assumes normal distribution)

this change:

Result "io.quarkus.arc.crazybeans.test.ArcBuildTimeMeasurementJmhBenchmark.benchmarkArCInitialization":
  0.213 ±(99.9%) 0.002 s/op [Average]
  (min, avg, max) = (0.210, 0.213, 0.215), stdev = 0.001
  CI (99.9%): [0.211, 0.215] (assumes normal distribution)


dmlloyd:
I'm beginning to wonder if we should also record some GC statistics as well. If these changes result in improvement due to fewer GC cycles, it would explain how these numbers are coming to be.

gsmet:
Fwiw, in the unfiltered JFR files, there is a lot of GC activity. 

I can get some data if you tell me what you want to collect.

dmlloyd (approved):

PR #504, opened 2025-08-27
Try an alternative approach to caching SAM location

dmlloyd:
This should avoid class loader leaks, since the value being cached is a JDK class.

### DIFF
--- a/src/main/java/io/quarkus/gizmo2/impl/Util.java
+++ b/src/main/java/io/quarkus/gizmo2/impl/Util.java
@@ -221,14 +221,8 @@ public final class Util {
         }
     }
 
-    // I'm not exactly sure why, but the following usage of `ClassValue` seems to cause a classloader leak.
-    // It feels likely the cause is more subtle, or entirely different, but avoiding `ClassValue` caching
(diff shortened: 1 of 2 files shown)


src/main/java/io/quarkus/gizmo2/impl/Util.java
```diff
@@ -243,36 +237,14 @@ protected MethodDesc computeValue(final Class<?> type) {
                 }
             }
             if (sam == null) {
-                throw new IllegalArgumentException("No abstract method found on " + type);
+                throw new IllegalArgumentException("No SAM found on " + type);
```
Ladicek, line 240: BTW, I originally changed the wording from "SAM" to "abstract method" because:

1. "SAM" is a Java/JVM engineer colloquialism that doesn't necessarily have to be well known to ordinary population.
2. There are 2 reasons for why you cannot find a SAM: there's no abstract method, or there's more than one abstract method. Since we throw different error messages in these situations, I find it appropriate to mention "abstract method" in both of them (or none, but why just in one and not the other?).

Ladicek (approved):

dmlloyd:
Oops! I guess I forgot the `^`

PR #507, opened 2025-09-03
remove `StringBuilderOps` and document reusability of `Ops` instances

Ladicek:
The `StringBuilderOps` class is removed due to disputes over the API design.

Futrher, this commit clarifies that instances of all `*Ops` classes are _not_ reusable: only one method may be called on the result of `BlockCreator.with*()`.


dmlloyd:
> > Java is not a valid parallel here. In Java you would have to assign to an intermediate variable in this case. The usage in Gizmo would be _exactly the same_ and in line with a principle we already have: if you would assign the value to a variable in Java, then you would assign the value to a variable in Gizmo as well. Consider:
> > ```java
> > StringBuilder b = new StringBuilder().append("foo").append("bar");
> > if (condition) {
> >     b.append("zab");
> > } else {
> >     b.append("baz");
> > }
> > b.append("oof");
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > is equivalent to:
> > ```java
> > Var b = b1.localVar("b", b1.withNewStringBuilder().append("foo").append("bar"));
> > b1.if_(condition, b2 -> {
> >     b2.withStringBuilder(b).append("zab");
> > }, b2 -> {
> >     b2.withStringBuilder(b).append("baz");
> > });
> > b1.withStringBuilder(b).append("oof");
> > ```
> 
> Right, but that's not the kind of condition I had in mind. It briefly occurred to me that there are, for lack of better terms, conditions (conditions in the generated code) and meta-conditions (conditions in the _generating_ code), but I didn't expand on it -- sorry. I had the 2nd kind in mind.

I see. So a more accurate example would be:

```java
Var b = b1.localVar("b", b1.withNewStringBuilder().append("foo").append("bar"));
if (condition) {
    b1.withStringBuilder(b).append("zab");
} else {
    b1.withStringBuilder(b).append("baz");
}
b1.withStringBuilder(b).append("oof");
```

> > Like any expression - you should not reuse an `Ops` instance. It does not for example track nested block state, so if you reused it within the `if` branches, you'd get an exception.
> 
> OK, I think I finally see where you're coming from. You treat `Ops` as expressions -- I don't, because they don't implement `Expr`. I guess it makes sense though, from certain POV.
> 
> > The absolute last thing we want is for users to ask "can I reuse this value?" and have the answer be "sometimes". The answer should always be "vars or constants, yes; expressions, no".
> 
> OK, so `withNewStringBuilder()` allocates a new local variable as documented. I don't see a way around that. `withStringBuilder()` accepts an `Expr` and allocates a local variable if the argument is not a `Var` already. I have no problem with making `withStringBuilder()` accept a `Var` instead.
> 
> > We also want to avoid overproliferation of local variables, which is one of the problems with Gizmo 1 already. An `Ops` is nothing more or less than a wrapper over its input, so the answer transitively applies to the wrapper instance as well.
> 
> OK, this makes perfect sense to me. So if a `with*()` method (and the constructor) accepts an `Expr`, the `Ops` instance is not reusable. If the `with*()` method (and the constructor) accepts a `Var`, the `Ops` instance is reusable. Agree?

It is reusable *within the same exact block*. It seems to me that this can be fairly error-prone though. The point of creating these instances off of the block is that they are only good within that exact point in the code. If we want some kind of Java-fluent-style thing then it seems to me we'd want to invert it so that the view accepts the block instead of vice-versa:

```java
StringBuilderOps sbView = StringBuilderOps.of(sb);
sbView.append(b1, "foo");
b1.loop_(b2 -> {
    sbView.append(b2, "bar");
});
```

or maybe:

```java
var sb = b1.declare("sb", StringBuilder.class).withStringBuilderOps();
sb.append(b1, "foo");
b1.loop_(b2 -> {
    sb.append(b2, "bar");
}
```

or maybe:

```java
StringBuilderOps sbView = StringBuilderOps.over(sb);
b1.with(sbView).append("foo");
b1.loop_(b2 -> {
     b2.with(sbView).append("foo");
});
```

or something like that.

Then your statement would be strictly true, that the view has the same reusability as the expression it wraps.

The fundamental disconnect is that we're *building* code not *executing* it, and all operations are in terms of creators. The `Ops` really only make sense when looked at this way; otherwise, you basically have two different/opposed models fighting each other, which is never a good thing when designing an API.

One other thing to mention is that if there is a `newStringBuilder` method which creates a view over a variable that is created on the fly, then it should probably accept a variable name.

Ladicek:
Just pushed a small change to illustrate what I have in mind at the moment.

Ladicek:
With the recent change in #522, the `StringBuilderOps` being created on a `Var` is no longer error-prone to misuse across different `BlockCreator`s. Use from a different block will result in an exception, just like in all other cases.

PR #522, opened 2025-09-22
more complete verification of calls on enclosing `BlockCreator`s

Ladicek:
We already had a facility to verify that a `BlockCreator` whose "child" is currently being created cannot be used. This facility was very limited though -- it only applied in the `block()` and `blockExpr()` methods. This commit expands that to all methods that contribute to the same method, such as conditions, loops or exception handlers. A test is added that verifies all these.

Note that we do _not_ attempt to verify this across multiple methods, including lambdas and anonymous classes. This would be more costly.


src/main/java/io/quarkus/gizmo2/impl/Util.java
```diff
@@ -151,7 +151,9 @@ public static String trackCaller() {
 
     private static String callerOutsideGizmo() {
         return SW.walk(stream -> stream
-                .filter(it -> !it.getClassName().startsWith("io.quarkus.gizmo2") || it.getClassName().endsWith("Test"))
+                .filter(it -> !it.getClassName().startsWith("io.quarkus.gizmo2")
```
dmlloyd, line 154: I didn't think of this before, but it might be slightly more robust to say `! it.getPackageName().equals(...)`. You could then use a `Set` to validate the package name if you wanted to simplify it even further.

dmlloyd (approved):
I like the change overall. It does introduce a smidge more allocation but I think that's OK.

src/main/java/io/quarkus/gizmo2/impl/Util.java
Ladicek: `StackFrame` doesn't have a `getPackageName()` method. I could use `getDeclaringClass().getPackageName()`, which seems like a longer way of writing what's currently there, but I don't think I could put all package names into a `Set`, because I have to check subpackages as well.

dmlloyd: Ah, yes you're right. I was thinking of it as `Class`.

PR #528, opened 2025-09-30
improvements

Ladicek:
- add ability to emit lambdas as anonymous classes
- avoid declaring local variables when debug info is disabled
  - This one is particularly interesting. The previous solution (which I suggested) indeed avoids emitting the LVT, but it doesn't avoid emitting the (otherwise useless) entries in the constant pool. So I had to add what @dmlloyd already previously had.
- fix emitting nests

### DIFF
--- a/src/main/java/io/quarkus/gizmo2/Gizmo.java
+++ b/src/main/java/io/quarkus/gizmo2/Gizmo.java
@@ -30,6 +30,8 @@ public sealed interface Gizmo permits GizmoImpl {
 
     /**
      * {@return a Gizmo instance which has debug info output enabled or disabled}
+     * <p>
+     * By default, debug info is included.
      *
      * @param debugInfo {@code true} to include debug info, or {@code false} to exclude debug info
      */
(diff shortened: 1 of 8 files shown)


src/main/java/io/quarkus/gizmo2/impl/BlockCreatorImpl.java
```diff
@@ -702,6 +711,7 @@ public Expr newAnonymousClass(final ConstructorDesc superCtor, final List<? exte
         List<MethodModel> methods = cm.methods();
         MethodModel ourCtor = methods.get(methods.size() - 1);
         owner.output().write(desc, bytes);
+        owner.addNestMember(desc);
```
dmlloyd, line 714: Oops, good catch

src/main/java/io/quarkus/gizmo2/Gizmo.java
```diff
@@ -40,11 +42,28 @@ static Gizmo create(ClassOutput outputHandler) {
...
+     * but through regular instantiation of the additional anonymous classes.
+     * <p>
+     * This doesn't affect runtime semantics of generated code, but it does affect the number
+     * of generated classes and, as of this writing, also runtime memory consumption.
+     * <p>
+     * By default, lambdas are <em>not</em> generated as anonymous classes.
+     *
+     * @param lambdasAsAnonymousClasses whether lambdas should be generated as anonymous classes
+     */
+    Gizmo withLambdasAsAnonymousClasses(boolean lambdasAsAnonymousClasses);
```
gsmet, line 65: Let me ask a dumb question: would it be possible to determine if we can generate a separate class instead when possible (the equivalent of a static nested class) from the lambda?

Asking because then the name might not be ideal. E.g. I could see how this method could be used to generate separate classes implementing Function/BiFunction/... or fall back to an anonymous class if not possible.

Question comes from the discussions we had with the anonymous classes being more verbose than the previous setup generating separate named classes.

dmlloyd (approved):
LGTM. I wonder if it's worth filing an upstream bug about the constant pool.

src/main/java/io/quarkus/gizmo2/Gizmo.java
dmlloyd: To be able to use an unrelated class, the lambda method which implements the behavior would have to be public, or else it would have to be moved to the implementation class (with the side-effect that then, all of the things the "lambda" *accesses* would then have to be public).

In the absence of making things public, it's not clear to me that this would work for *most* cases (it'd probably work for a few cases though, like pure functions (comparators etc.)).

Ladicek: I _guess_ you're thinking about a more elaborate scheme where we would still generate the lambda body into a special synthetic method, but then also generate a single class for each shape of a lambda (the implemented interface + maybe the list of types of captured variables) which would then call the lambda body somehow? That's not what this PR does and I don't think we can evaluate the resource usage of that scheme without implementing it. If anyone wants to experiment around that, feel free :-)

For a description of what this PR does, see the commit message, but I'll also expand on it here.

In the default mode:

- a lambda body is generated into a special synthetic method in the class we're currently generating,
- captured variables become extra parameters of this method
- a lambda instance is obtained by `invokedynamic` whose bootstrap method is `LambdaMetafactory.metafactory()`

But in the "lambdas as anonymous classes" mode:

- a lambda body is generated into a separate anonymous class (which is a member of the nest "owned" by the generated class, so visibility doesn't matter)
- captured variables become fields of this anonymous class
- a lambda instance is obtained by creating an instance of this anonymous class

For each lambda, we generate a new anonymous class. If we generate a lambda for `Function`, the anonymous class implements `Function`, etc.

> Question comes from the discussions we had with the anonymous classes being more verbose than the previous setup generating separate named classes.

They are indeed more verbose, but also take _less_ memory at runtime than "proper" lambdas. That's why I added this option.

PR #533, opened 2025-10-06
Compute generic types only when needed

### LINKED ISSUE #490: Should we have a small cache of `GenericType`?
### Issue description

I gathered the calls to `GenericType.of` and dropped the ones that are specific to ArC.

Note that some are probably still that high because of very specific ArC things such as `Supplier` but I think it makes for a good list of what could be usual suspects.

Note that there are two options here: we just add a CHM with a cache or we specifically cache the most common types and let the others be instantiated every time.

```
   2402 GenericType.of: ClassOrInterfaceDesc[HashSet]
  32706 GenericType.of: ClassOrInterfaceDesc[Map]
  46262 GenericType.of: ClassOrInterfaceDesc[ClassLoader]
  46262 GenericType.of: ClassOrInterfaceDesc[Thread]
  62000 GenericType.of: ArrayClassDesc[Class, 1]
  62000 GenericType.of: ArrayClassDesc[Object, 1]
  62000 GenericType.of: ClassOrInterfaceDesc[BiFunction]
  62190 GenericType.of: ClassOrInterfaceDesc[HashMap]
  85883 GenericType.of: ClassOrInterfaceDesc[List]
  92655 GenericType.of: ClassOrInterfaceDesc[Exception]
 123308 GenericType.of: ClassOrInterfaceDesc[RuntimeException]
 131536 GenericType.of: class java.lang.Object
 140781 GenericType.of: ClassOrInterfaceDesc[Type]
 170286 GenericType.of: void
 184890 GenericType.of: ClassOrInterfaceDesc[String]
 224048 GenericType.of: ClassOrInterfaceDesc[Class]
 231179 GenericType.of: ClassOrInterfaceDesc[Set]
 275914 GenericType.of: ClassOrInterfaceDesc[Supplier]
1642212 GenericType.of: ClassOrInterfaceDesc[Object]
```

dmlloyd:
Reduce `GenericType` usage to only cases where they are required. Fixes #490.

This only ended up taking me an hour or so, so I guess I was worried over nothing.

### DIFF
--- a/src/main/java/io/quarkus/gizmo2/Expr.java
+++ b/src/main/java/io/quarkus/gizmo2/Expr.java
@@ -17,10 +17,15 @@ public sealed interface Expr extends SimpleTyped permits Const, Assignable, This
 
     /**
      * {@return the generic expression type (not {@code null})}
+     * If {@link #hasGenericType()} returns {@code false},
+     * the returned type will be a wrapper around {@link Typed#type() type()}.
      */
-    default GenericType genericType() {
-        return GenericType.of(type());
-    }
+    GenericType genericType();
+
+    /**
+     * {@return {@code true} if this value has a generic type, or {@code false} if it does not have one}
+     */
+    boolean hasGenericType();
 
     /**
      * {@return an assignable for an element of this array}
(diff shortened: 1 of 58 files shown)


Ladicek (approved):
Code changes LGTM, though I'd (once again :-) ) like to highlight that the amount of code to deal with generic types greatly outweigh its usefulness (in the whole ArC code base, there's _one_ place where we generate generic types, and we do that incorrectly). I'd just get rid of this, with exception of declarations (like classes, methods and fields) where it makes sense. But I realize I'm a minority here :-)

dmlloyd:
I'm thinking a bit bigger and longer term. The entire point of Gizmo 2 to begin with is to make something more powerful and flexible. The consequence of this is not only to make our existing code nicer, but also to expand into new use cases as well. ArC is a nice test bed to validate what we have, but it does not represent the limit of the usefulness of Gizmo, nor does it represent the boundaries of reasonable use cases.

PR #556, opened 2025-11-17
Performance improvements

dmlloyd:
Changes to reduce redundant allocations and avoid the somewhat heavy `ClassDesc#hashCode` and `ClassDesc#equals` methods.


dmlloyd:
@gsmet reported around 5% improvement on Java 17 on a single test.

PR #569, opened 2025-11-19
Avoid hitting slow map for common class descriptors

dmlloyd:
When accepting `Class<?>` arguments, try to avoid hitting the slow synchronized map behind `ClassValue` for common types. The common types are gleaned from our known constant fields. We're trading a bit of class init time for higher throughput. An alternative would be to hand-write the map, which is a little more error-prone but could possibly be faster than using reflection.


dmlloyd:
I want to ponder this a bit more and see if I can make it have a measurable performance benefit in some form.

PR #651, opened 2026-06-23
Add nested/inner class/interface support

### LINKED ISSUE #383: Support for nested classes
### Issue description

We have incomplete support for nested classes; we should round it out a bit better and support both inner and static nested classes.

- [ ] Inner class support
- [ ] Static nested class support

dmlloyd:
Fixes #383


src/main/java/io/quarkus/gizmo2/creator/TypeCreator.java
```diff
@@ -181,6 +181,30 @@ default StaticFieldVar constantField(String name, Const value) {
...
+     * by whether the {@link ModifierFlag#STATIC STATIC} flag is set by the builder.
+     * <p>
+     * For non-static inner classes, a synthetic {@code this$N} field and constructor parameter
+     * are automatically injected to capture the enclosing instance.
+     *
+     * @param simpleName the simple name of the nested class (must not be {@code null})
+     * @param builder the builder for the nested class (must not be {@code null})
+     * @return the descriptor of the nested class (not {@code null})
+     */
+    ClassDesc nestedClass(String simpleName, Consumer<ClassCreator> builder);
```
Ladicek: I think what you can create through these methods is akin to:

```java
class Outer {
    [static] class Inner {
    }
}
```

Correct?

In other words, you cannot create a local or anonymous class through these methods.

If so, I think the best name is not "nested class" but "member class", per https://docs.oracle.com/javase/specs/jls/se25/html/jls-8.html#jls-8.5

Ladicek: Forgot to mention: https://docs.oracle.com/javase/specs/jls/se25/html/jls-8.html says that

> A _nested class_ is any class whose declaration occurs within the body of another class or interface declaration. A nested class may be a member class ([§8.5](https://docs.oracle.com/javase/specs/jls/se25/html/jls-8.html#jls-8.5), [§9.5](https://docs.oracle.com/javase/specs/jls/se25/html/jls-9.html#jls-9.5)), a local class ([§14.3](https://docs.oracle.com/javase/specs/jls/se25/html/jls-14.html#jls-14.3)), or an anonymous class ([§15.9.5](https://docs.oracle.com/javase/specs/jls/se25/html/jls-15.html#jls-15.9.5)).
>
> A nested class may be _inner_ ([§8.1.3](https://docs.oracle.com/javase/specs/jls/se25/html/jls-8.html#jls-8.1.3)), in which case it may (depending on precisely where its declaration occurs) be able to refer to enclosing class instances, local variables, and type variables.

src/main/java/io/quarkus/gizmo2/creator/TypeCreator.java
```diff
@@ -181,6 +181,30 @@ default StaticFieldVar constantField(String name, Const value) {
...
+
+    /**
+     * Create a named nested interface within this type.
+     * Nested interfaces are always implicitly static.
+     *
+     * @param simpleName the simple name of the nested interface (must not be {@code null})
+     * @param builder the builder for the nested interface (must not be {@code null})
+     * @return the descriptor of the nested interface (not {@code null})
+     */
+    ClassDesc nestedInterface(String simpleName, Consumer<InterfaceCreator> builder);
```
Ladicek: Same here, it's a member interface, not nested interface.

https://docs.oracle.com/javase/specs/jls/se25/html/jls-9.html says that

> A _nested interface_ is any interface whose declaration occurs within the body of another class or interface declaration. A nested interface may be a member interface ([§8.5](https://docs.oracle.com/javase/specs/jls/se25/html/jls-8.html#jls-8.5), [§9.5](https://docs.oracle.com/javase/specs/jls/se25/html/jls-9.html#jls-9.5)) or a local interface ([§14.3](https://docs.oracle.com/javase/specs/jls/se25/html/jls-14.html#jls-14.3)).

src/main/java/io/quarkus/gizmo2/creator/TypeCreator.java
dmlloyd: It would probably make the most sense to just call this method `class_` like we do at the top level. Then the kind of class pretty much just depends on the context in which the method was called. If we ever add local class support, then it would just be a `class_` method inside of a `BlockCreator` I guess (not that there's a whole lot of demand to be able to create local classes).

dmlloyd:
Fixed inaccurate member class naming.

src/main/java/io/quarkus/gizmo2/creator/TypeCreator.java
Ladicek: `class_` (and `interface_` on the other place) makes sense to me.

src/test/java/io/quarkus/gizmo2/MemberClassTest.java
```diff
@@ -0,0 +1,518 @@
...
+                    ctc.body(b0 -> {
+                        b0.invokeSpecial(ConstructorDesc.of(Object.class), ic.this_());
+                        b0.return_();
+                    });
+                });
+                ic.method("getOuterValue", mc -> {
+                    mc.returning(Object.class);
+                    mc.body(b0 -> {
+                        // access outer's value via this$0
+                        FieldDesc this0 = FieldDesc.of(innerDesc, "this$0", outerDesc);
```
Ladicek, line 110: Should we provide an accessor for the enclosing instance on the member class creator? We don't have an interface for it at the moment, but maybe we should?

src/test/java/io/quarkus/gizmo2/MemberClassTest.java
```diff
@@ -0,0 +1,518 @@
...
+                mc.returning(Object.class);
+                mc.body(b0 -> {
+                    b0.return_(Const.of("hello from member interface impl"));
+                });
+            });
+            cc.staticMethod("test", mc -> {
+                mc.returning(Object.class);
+                mc.body(b0 -> {
+                    Expr outer = b0.new_(ConstructorDesc.of(outerDesc));
+                    b0.return_(b0.invokeVirtual(ClassMethodDesc.of(outerDesc, "greet", Object.class), outer));
```
Ladicek: Do you actually want an `invokeVirtual` here? I would expect `invokeInterface` through the member interface.

src/test/java/io/quarkus/gizmo2/MemberClassTest.java
Ladicek: Alternatively, but outside of current Gizmo convention, it doesn't have to be on the member class creator, but as a second lambda parameter of the member class consumer. I was thinking we could do this for `this` on top-level classes as well, but as mentioned above, it would be outside current Gizmo convention.

Ladicek (approved):
LGTM, just one idea and one open question about a test.

src/test/java/io/quarkus/gizmo2/MemberClassTest.java
dmlloyd: The intention is that since this is strictly nested in the enclosing class, you can use the captured `this` from the enclosing type directly.

Ladicek: That would make perfect sense, but it's clearly not what you're doing here :-)

dmlloyd: Fixed; now we recursively search outer scopes, chaining `this$N` reads until we find the target scope (or fail trying).

src/test/java/io/quarkus/gizmo2/MemberClassTest.java
dmlloyd: Fixed. I'll blame Claude for this one. :-)

src/test/java/io/quarkus/gizmo2/MemberClassTest.java
Ladicek: GitHub might be lying to me, but I still see

```java
FieldDesc this0 = FieldDesc.of(innerDesc, "this$0", outerDesc);
Expr outerRef = ic.this_().field(this0);
b0.return_(outerRef.field(outerField));
```

here, where I would expect

```java
b0.return_(cc.this_().field(outerField));
```

src/test/java/io/quarkus/gizmo2/MemberClassTest.java
```diff
@@ -0,0 +1,697 @@
...
+                    ctc.body(b0 -> {
+                        b0.invokeSpecial(ConstructorDesc.of(Object.class), ic.this_());
+                        b0.set(ic.this_().field(innerField), label);
+                        b0.return_();
+                    });
+                });
+                ic.method("describe", mc -> {
+                    mc.returning(Object.class);
+                    mc.body(b0 -> {
+                        FieldDesc this0 = FieldDesc.of(innerDesc, "this$0", outerDesc);
```
Ladicek, line 226: This place could/should also use proper `this` support.

src/test/java/io/quarkus/gizmo2/MemberClassTest.java
Ladicek: Ah, please ignore, I just noticed there's a new test that does check this. So it's OK.

src/test/java/io/quarkus/gizmo2/MemberClassTest.java
Ladicek: Ah, please ignore, I just noticed there's a new test that does check this. So it's OK.

PR #653, opened 2026-06-25
Source generation support

### LINKED ISSUE #419: add a feature to dump source code
### Issue description

This is similar to what Gizmo 1 had and we should use the same API (or at least as close as possible). The format doesn't need to be the same; in fact, I think it would be best if the format was as close to the Java language as possible. That should be doable most of the time, except of things like `invokedynamic`.

This is for 2.1, does not block 2.0 in any way IMHO.

dmlloyd:
Includes line number correlation.

Fixes #419

### DIFF
--- a/src/main/java/io/quarkus/gizmo2/impl/constant/DoubleConst.java
+++ b/src/main/java/io/quarkus/gizmo2/impl/constant/DoubleConst.java
@@ -4,6 +4,7 @@ import java.lang.constant.ConstantDescs;
 import java.util.Optional;
 
 import io.quarkus.gizmo2.impl.BlockCreatorImpl;
+import io.quarkus.gizmo2.impl.SourceBuilder;
 import io.quarkus.gizmo2.impl.StackMapBuilder;
 import io.smallrye.classfile.CodeBuilder;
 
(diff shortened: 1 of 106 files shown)


src/main/java/io/quarkus/gizmo2/impl/constant/DoubleConst.java
```diff
@@ -112,4 +113,10 @@ public Optional<Double> describeConstable() {
     public StringBuilder toShortString(final StringBuilder b) {
         return b.append(value).append(" (").append(Double.toHexString(value)).append(')');
     }
+
+    /** {@inheritDoc} */
+    @Override
+    protected StringBuilder appendSourceExpr(StringBuilder buf, SourceBuilder sb) {
+        return buf.append(desc()).append('d');
```
Ladicek: I think idiomatic Java doesn't append the `d` suffix for `double` values. Doesn't really matter.

Ladicek (approved):
LGTM, although I admit this was really a cursory review. I would have expected most `Item`s to not be changed at all, maybe to expose some necessary information, and the entire formatting would be a big pattern matching `switch`, but I don't mind this either.

dmlloyd:
It actually started off that way... well, as an if/else tree actually, since the target JDK was 17. But, even pattern matching is still a linear search IIRC so it rubbed me the wrong way.

PR #654, opened 2026-06-29
Provide a way to set the version at the Gizmo level

gsmet:
We plan to set the minimum for Quarkus to Java 21, but I think it makes sense to keep the default for Gizmo to Java 17 for now - except if at some point we have a strong reason to move to 21.

This PR introduces a more practical way to set the bytecode version right at the Gizmo instance level.


Ladicek (approved):
LGTM, though I don't understand why the call to `zb.withVersion()` was moved from `preAccept()` to `postAccept()`, but that probably doesn't make a difference.

I'll defer to @dmlloyd whether he's fine with this.

gsmet:
`ClassVersionTest.perClassSetVersionOverridesGlobal` is failing if not.

We need to write the version after the version has been set in the user's callback, not before.

dmlloyd (approved):

